PR #14, opened 2022-09-13
feat: support all types of attributes

makeavish:


srikanthccv (commented):
We should do the same for resource attributes here https://github.com/SigNoz/signoz-otel-collector/blob/main/exporter/clickhousetracesexporter/clickhouse_exporter.go#L244?

nityanandagohain (dismissed):

ankitnayan (approved):

PR #16, opened 2022-09-23
fix: skip spans with negative duration

srikanthccv:


ankitnayan (approved):

makeavish:
Why don't we use `uint64` for latency, we are doing the same for traces

srikanthccv:
The traces logic suffers from unsigned integer overflow issues. Try setting the end time before start and see what does `durationNano` contain.

makeavish:
Oh, I assumed negative integers are converted to 0 when converting `uint64`.
Will update traces logic to set durationNano zero for negative cases, should we do same in metrics?

srikanthccv:
I believe we are not doing that anywhere else in the metrics. This was the only place and we now skip such spans from including in the metrics.

makeavish:
@ankitnayan What would you suggest? Should we convert spans with negative duration to zero duration for traces and metrics?

srikanthccv:
I think we should convert and keep zero value for spans otherwise it produces an incorrect result for any calc made on the `durationNano` column. We can't also skip because it breaks the trace.

PR #17, opened 2022-09-28
fix: override repeated label keys

srikanthccv:
Fixes https://github.com/sigNoz/engineering-pod/issues/727

### DIFF
--- a/exporter/clickhousemetricsexporter/clickhouse.go
+++ b/exporter/clickhousemetricsexporter/clickhouse.go
@@ -232,15 +232,19 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest) erro
 
 	for i, ts := range data.Timeseries {
 		var metricName string
-		labels := make([]*prompb.Label, len(ts.Labels))
-		for j, label := range ts.Labels {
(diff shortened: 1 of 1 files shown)


nityanandagohain (approved):

ankitnayan (approved):

prashant-shahi:
@srikanthccv it would have been nice to have an `override` flag. If set to `true`, it would override or else skip in case of repeated labels.

srikanthccv:
When we say skip it leads to follow up questions which one to skip? Should it be most recent or last recent or should it pick something middle? I think this is unknown until one looks at the values and see which one is desired. 

This is inline with what SDK's already do today and what spec recommends in case of duplicates.

https://github.com/open-telemetry/opentelemetry-go/blob/b369e59ba15baa5d6bd0304bf3ea4e7a26f301fe/sdk/trace/span.go#L208-L209

> Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value

https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common#attribute-collections

prashant-shahi:
Got it! It makes sense to adhere with the specification.

PR #22, opened 2022-10-20
Usage collection via otel collector

nityanandagohain:
This PR adds the changes to collect usage metrics from otel collector.

FIXES https://github.com/SigNoz/engineering-pod/issues/718


usage/usage.go
```diff
@@ -0,0 +1,135 @@
...
+	tableName      string
+	usageParser    func(metrics []*metricdata.Metric) (map[string]Usage, error)
+	prevCount      int64
+	prevSize       int64
+}
+
+var CollectorID uuid.UUID
+
+func init() {
+	CollectorID = uuid.New()
```
ankitnayan, line 45: @nityanandagohain should we use existing otel-collector id rather than creating a new one. By default, an otel-collector id is created which is called service-instance-id which can be reused

usage/usage.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func NewUsageCollector(db clickhouse.Conn, options Options, dbName string, usageParser func(metrics []*metricdata.Metric) (map[string]Usage, error)) *UsageCollector {
+	return &UsageCollector{
+		id:          uuid.New(),
+		reader:      metricexport.NewReader(),
+		o:           options,
+		db:          db,
+		dbName:      dbName,
+		tableName:   "usage",
```
ankitnayan, line 55: @nityanandagohain we should define the table name somewhere as constant.

usage/usage.go
```diff
@@ -0,0 +1,135 @@
...
+		usageParser: usageParser,
+		prevCount:   0,
+		prevSize:    0,
+	}
+}
+func (e *UsageCollector) CreateTable(db clickhouse.Conn, databaseName string) error {
+	query := fmt.Sprintf(
+		`
+		CREATE TABLE IF NOT EXISTS %s.usage (
```
ankitnayan, line 64: @nityanandagohain we should define the table name somewhere as constant.

usage/usage.go
nityanandagohain: Wasn't able to get a way to extract the collector ID that is already allocated. Will try again.

nityanandagohain:
PR is updated now to support distributed clickhouse.

ankitnayan (approved):

PR #28, opened 2022-11-08
Add some additional components

srikanthccv:
Add some new components that are in the alpha/dev stage.


cmd/signozcollector/components.go
```diff
@@ -42,6 +43,8 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourcedetectionprocessor"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourceprocessor"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/routingprocessor"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/schemaprocessor"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/servicegraphprocessor"
```
srikanthccv, line 47: We do a better job than this processor in terms of recognizing the services in some cases, but where I imagine this will benefit is similar to the span metrics processor, where the users run these processors for collecting aggregate and send that info and can drop spans in their pipeline before sending trace data to us because they want to avoid the n/w costs.

ankitnayan (approved):

PR #34, opened 2022-11-13
Remove labels_object column from time_series_v2

srikanthccv:


exporter/clickhousemetricsexporter/clickhouse.go
```diff
@@ -94,17 +94,12 @@ func NewClickHouse(params *ClickHouseParams) (base.Storage, error) {
...
-	// in clickhouse-go. We workaround this limitation for now by
-	// using the DEFAULT expression. However, we can use labels_object
-	// in the querying for faster results.
 	queries = append(queries, fmt.Sprintf(`
 		CREATE TABLE IF NOT EXISTS %s.time_series_v2 (
 			metric_name LowCardinality(String),
 			fingerprint UInt64 Codec(DoubleDelta, LZ4),
 			timestamp_ms Int64 Codec(DoubleDelta, LZ4),
 			labels String Codec(ZSTD(5)),
-			labels_object JSON DEFAULT labels CODEC(ZSTD(5))
```
ankitnayan, line 107: @srikanthccv this won't drop the column for existing users and they would face issues, right? IMO we should have a query like `alter table.... drop column if exists ....`

srikanthccv: They won't face any issues because this uses `DEFAULT`, but it will continue to exist, which is unnecessary because it will not be used. I will add the alter query as well.

srikanthccv: Done. I will create an issue for myself and move these to migration files as we do for traces and logs.

ankitnayan (approved):

PR #47, opened 2022-12-14
Update span metrics processor

srikanthccv:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/21

How to verify wonder issue is fixed:

Run the v0.12.0 version and set the small [`dimensions_cache size`](https://github.com/SigNoz/signoz/blob/develop/deploy/docker/clickhouse-setup/otel-collector-config.yaml#L77), such as 1, to trigger the issue and notice the services tab and traces with the count. You will see that traces are dropped without exporting when the error happens. 

Now run the same setup with this fix -- the traces should be exported.


srikanthccv:
Tests are failing since this PR https://github.com/SigNoz/signoz-otel-collector/pull/22 introduced a change that requires a ClickHouse connection. It will be addressed separately.

ankitnayan:
@srikanthccv did updating the `cache.go` to the latest fixed the issue automatically or did you make changes specific to signoz? I only see major changes in the `cache.go` file

ankitnayan (approved):

processor/signozspanmetricsprocessor/processor.go
```diff
@@ -273,81 +298,90 @@ func (p *processorImp) Capabilities() consumer.Capabilities {
 // It aggregates the trace data to generate metrics, forwarding these metrics to the discovered metrics exporter.
 // The original input trace data will be forwarded to the next consumer, unmodified.
 func (p *processorImp) ConsumeTraces(ctx context.Context, traces ptrace.Traces) error {
-	// p.logger.Info("Consuming trace data")
+	// Forward trace data unmodified and propagate both metrics and trace pipeline errors, if any.
+	return multierr.Combine(p.tracesToMetrics(ctx, traces), p.nextConsumer.ConsumeTraces(ctx, traces))
+}
+
+func (p *processorImp) tracesToMetrics(ctx context.Context, traces ptrace.Traces) error {
```
srikanthccv, line 305: @ankitnayan this diff fixes the wonder issues since next consumer gets data even when there is an error with traceToMetrics function

processor/signozspanmetricsprocessor/processor.go
```diff
@@ -15,88 +15,120 @@
...
+	callLatencyBounds []float64
 
-	latencyBounds []float64
+	dbCallHistograms    map[metricKey]*histogramData // signoz_db_latency_* metric
+	dbCallLatencyBounds []float64
 
-	dbLatencyCount map[metricKey]uint64
-	dbLatencySum   map[metricKey]float64
+	externalCallHistograms    map[metricKey]*histogramData // signoz_external_call_latency_* metric
+	externalCallLatencyBounds []float64
```
ankitnayan, line 95: we don't need percentiles of external calls, we are just showing `avg` duration in the charts. It was causing a lot of high cardinality data and hence I remove buckets from externalCalls earlier

processor/signozspanmetricsprocessor/processor.go
ankitnayan: If this is a custom fix then the upstream spanmetricsprocessor would also stop forwarding traces to the next consumer on getting errors

processor/signozspanmetricsprocessor/processor.go
srikanthccv: That is still the same behaviour; please take a look at `collectExternalCallMetrics` function. I left it here just in case we need it.

processor/signozspanmetricsprocessor/processor.go
srikanthccv: It was also an issue in the upstream processor, and it's the same fix we are applying.

ankitnayan: ok

processor/signozspanmetricsprocessor/processor.go
ankitnayan: ok

PR #50, opened 2022-12-21
Wrapped collector

srikanthccv:

This adds a new wrapped collector component that adds the restart capability. The upstream [`service.Collector`](https://pkg.go.dev/go.opentelemetry.io/collector@v0.66.0/service#pkg-index) provides `Run`, `Shutdown`, and `GetState`. 

>Run starts the collector according to the given configuration, and waits for it to complete. Consecutive calls to Run are not allowed, Run shouldn't be called once a collector is shut down.

https://pkg.go.dev/go.opentelemetry.io/collector@v0.66.0/service#Collector.Run

For this reason, we create a new instance of Collector each time restart is triggered and don't allow Run to be called multiple times.

This will be used by the OpAMP client combined with the config manager to change the effective config dynamically.

Unlike `service.Collector` we don't wait for Run to wait for error and do not let it handle the SIGTERM and SIGINT on its own because we will be handling them in main.go and send the Shutdown call from there.

Handling all the failure modes is an important part of making this robust. Please review with a focus on that.


srikanthccv:
Following what's being advocated, I will send a PR first with interfaces/type definitions/etc...

ankitnayan:
Is it the major motive of the PR to make `service.Collector.Run()` asynchronous (non-blocking) to enable calls to the opAmp server? @srikanthccv

srikanthccv:
Yes, we also get certain control over the life cycle of the collector and how it runs (for instance its own signal handling for `SIGTERM`/`SIGINT` is disabled and we do the signal management).

ankitnayan:
> Yes, we also get certain control over the life cycle of the collector and how it runs (for instance its own signal handling for SIGTERM/SIGINT is disabled and we do the signal management [here](https://github.com/SigNoz/signoz-otel-collector/blob/10d0b83fbf21ab0a1d2aea9a726b232fd07c2b2f/cmd/signozcollector/main.go#LL62-L84)).

So, you plan otel-collector to respond to more signals from the opAMP client and hence extending? Not clear what are we gaining by controlling the lifecycle.  Is disabling the `col.set.ConfigProvider.Watch()` signal one of the use cases?

1. We are not respecting `DisableGracefulShutdown` anymore?
2. The set of signals the otel-collector honors are at https://github.com/open-telemetry/opentelemetry-collector/blob/release/v0.66.x/service/collector.go#L206-L232

srikanthccv:
>you plan otel-collector to respond to more signals from the opAMP client and hence extending? Not clear what are we gaining by controlling the lifecycle.

The OpAMP client manages when the collector should `Run`/`Shutdown`/`Restart`/`Rollback` which wouldn't be possible without 
this wrapper collector and controlling the life cycle. 

> Is disabling the col.set.ConfigProvider.Watch() signal one of the use cases?

This doesn't happen because we `Shutdown` first and then make file changes, and roll back appropriately.

>We are not respecting DisableGracefulShutdown anymore?

What this does is gracefully shut down all the exporters/processors/receivers. When we disable it, we are saying that it should not respond to SIGTERM/SIGINT on its own i.e Call `Shutdown`. This is done by the OpAMP client so it's still a graceful shutdown.

nityanandagohain (changes requested):
Have added a few comments for name changes in tests. Overall the changes look good to me.

nityanandagohain (approved):

mindhash (approved):

PR #51, opened 2022-12-21
Add simple client implementation

srikanthccv:

This adds a "simpleClient" implementation that runs the collector as it is done today. Collector, when not provided with the OpAMP server configuration to communicate, runs without any remote config capabilities.


nityanandagohain (approved):
LGTM! This can be reference for short independent PR's for us.

One thing I would prefer is keeping the test config files different for each of the client , ex:- testdata/simple/* and testdata/server/*

srikanthccv:
> One thing I would prefer is keeping the test config files different for each of the client , ex:- testdata/simple/* and testdata/server/*

👍

mindhash (approved):

nityanandagohain (approved):

PR #53, opened 2022-12-22
Add interfaces/types

srikanthccv:
This adds the interfaces and type definition required for implementing OpAMP with some detail on what each component is responsible for. Some rough flow diagram looks like following
```mermaid
flowchart LR
  subgraph Service
    Client
 end
 subgraph Client
    ConfigManager
    SigNozWrappedCollector
 end
 subgraph SigNozWrappedCollector
    otelcol.Collector 
 end
```

The main program interacts with `Service`, which runs the `Client`. There can be different implementations of client interfaces, currently serverClient (that connects and works with remote server) and simpleClient(no management as it is today). The serverClient implements the necessary callbacks for the remote config management https://pkg.go.dev/github.com/open-telemetry/opamp-go@v0.5.0/client/types#Callbacks. 

Pre-requisite for #50 and #51 and upcoming related PRs.


srikanthccv:
// @mindhash

mindhash:
@srikanthccv The config interactions seem to be in terms of [file data (bytes)
](https://github.com/open-telemetry/opamp-go/blob/cea1d9dc64099e9b68cf5fdd6ea2df90477e18fc/protobufs/opamp.pb.go#L2480).  

Just wondering, Is the following within the scope of this work - create config object (with changes from UI) in query-service (?). I think we could create a config struct and add methods that support addition of receivers, pipelines and processors.   If its common to all, there is better control.

srikanthccv:
Yes, we are trying to use the confmap provider from upstream that should suffice for most of the requirements https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap; if something is missing, we can extend. I will share some examples to make it more clear.

ankitnayan:
there seems to be a provision for partial updates from the provider. Example:
```
// `yaml:processors::batch::timeout: 2s`
// `yaml:processors::batch/foo::timeout: 3s`
```

Screenshot 2022-12-25 at 1 22 15 PM


But I don't see a URI that was updated as part of the change event from the provider and the [URI has to be provided by the Resolver ](https://github.com/open-telemetry/opentelemetry-collector/blob/main/confmap/resolver.go#L312) which is ideal if you know which part of conf you want to watch. This can work if we want to watch configs that signoz's UI is interested in. Eg, logs parer, sampling and similar configs. 

What I am not sure of is, when we watch for any changes and the resolver merges the changes, will it reload the complete config or will it reload only the parts changed?

srikanthccv:
>when we watch for any changes and the resolver merges the changes, will it reload the complete config or will it reload only the parts changed?

There is no partial reload. It restarts everything https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/collector.go#L187

srikanthccv:
And also, what we are aiming for is not to use this file watcher capability on the agent to effectively handle the rollbacks.

ankitnayan:
> > when we watch for any changes and the resolver merges the changes, will it reload the complete config or will it reload only the parts changed?
> 
> There is no partial reload. It restarts everything https://github.com/open-telemetry/opentelemetry-collector/blob/main/otelcol/collector.go#L187

Why is restarting needed?

ankitnayan:
and restarting will probably then create a new `service_instance_id`? So, a new timeseries will be creating every time a config changes. It will be like restarting a collector. IMO this is not a good way to handle updates.

ankitnayan:
> And also, what we are aiming for is not to use this file watcher capability on the agent to effectively handle the rollbacks.

Is rollback also the reason that adding a signoz provider [here](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider) isn't the best choice? I have not gone through all the PRs and thoughts, so this might be a trivial question.

opamp/config_manager.go
```diff
@@ -0,0 +1,42 @@
...
+	"go.uber.org/zap"
+)
+
+// agentConfigManager is responsible for managing the agent configuration
+// It is responsible for:
+// 1. Reading the agent configuration from the file
+// 2. Reloading the agent configuration when the file changes
+// 3. Providing the current agent configuration to the Opamp client
+type agentConfigManager struct {
+	agentConfig *remoteControlledConfig
```
ankitnayan, line 14: should this be locked to concurrent changes?

srikanthccv:
The ability to push down the partial updates to settings is not built into its components. This is generally the case across the project. I can bring it up sometime to see if this was a design decision that will not change.

srikanthccv:
>Is rollback also the reason that adding a signoz provider [here](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider) isn't the best choice? I have not gone through all the PRs and thoughts, so this might be a trivial question.

Yes

srikanthccv:
>and restarting will probably then create a new service_instance_id?

It is only regenerated if it's not provided in the config. I think this can be handled. I will play around and come back on this.

ankitnayan (dismissed):

ankitnayan:
> The ability to push down the partial updates to settings is not built into its components. This is generally the case across the project. I can bring it up sometime to see if this was a design decision that will not change.

I observed the same with Prometheus. I feel it's mostly because it was designed to work with files. For the same reasons, I introduced CRUD apis for the management of Prometheus alerts.

service/service.go
```diff
@@ -0,0 +1,38 @@
...
+	Start(ctx context.Context) error
+	Shutdown(ctx context.Context) error
+}
+
+type service struct {
+	l      *zap.Logger
+	client opamp.Client
+
+	managerConfigPath   string
+	collectorConfigPath string
```
ankitnayan, line 27: Do we need `managerConfigPath` and `collectorConfigPath` as part of the service struct?

srikanthccv: No, they are not needed here.

opamp/config_manager.go
ankitnayan: okay...you can mark the conv resolved when this is done

srikanthccv: responded to same question here https://github.com/SigNoz/signoz-otel-collector/pull/56#issuecomment-1364638488

srikanthccv:
@mindhash I re-read your question; I think I misunderstood it earlier. Did you mean something like this https://github.com/open-telemetry/opentelemetry-collector/blob/release/v0.66.x/service/unmarshaler.go#L37?

srikanthccv:
Please review so that all the relevant PRs can get in one by one. All of them can remain in the codebase independently until final glue up PR is merged.

mindhash (approved):
LGTM

nityanandagohain (approved):

PR #54, opened 2022-12-23
Add grok parser

nityanandagohain:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/55

PR for adding grok parser operator.

Please note:
The files cache.go and cache_test.go need not be reviewed as they are directly added from https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/stanza/operator/parser/regex without any changes.

I have created a pkg folder under which I have added this, please suggest an alternate structure if you feel this is not right.


nityanandagohain:
Without cache, for parsing the same kind of log

grok
```
BenchmarkParseNoCache-8                    	    4644	    256435 ns/op	  109221 B/op	    1204 allocs/op
```

regex
```
BenchmarkParseNoCache-8                    	    6739	    183266 ns/op	   92675 B/op	    1004 allocs/op
```

nityanandagohain:
There is a concurrency issue with this grok parser https://github.com/SigNoz/signoz-otel-collector/issues/66

pkg/parser/grok/grok.go
```diff
@@ -0,0 +1,164 @@
...
+
+// Build will build a grok parser operator.
+func (c Config) Build(logger *zap.SugaredLogger) (operator.Operator, error) {
+	parserOperator, err := c.ParserConfig.Build(logger)
+	if err != nil {
+		return nil, err
+	}
+
+	if c.Pattern == "" {
+		return nil, fmt.Errorf("missing required field 'grok'")
```
srikanthccv, line 59: there is no grok field in the config?
```suggestion
		return nil, fmt.Errorf("missing required field 'pattern'")
```

srikanthccv, line 59: there is no grok field in the config?
```suggestion
		return nil, fmt.Errorf("missing required field 'pattern'")
```

pkg/parser/grok/cache.go
```diff
@@ -0,0 +1,212 @@
...
+		limiter: newStartedAtomicLimiter(limit, interval),
+	}
+}
+
+// memoryCache is an in memory cache of items with a pre defined
+// max size. Memory's underlying storage is a map[string]item
+// and does not perform any manipulation of the data. Memory
+// is designed to be as fast as possible while being thread safe.
+// When the cache is full, new items will evict the oldest
+// item using a FIFO style queue.
```
srikanthccv, line 53: Any reason why this is used here instead of the LRU kind cache? Say there is an item which is added first to the queue, but it's also the most commonly used one for all the log messages, in that kicking that out in favour of some new item doesn't seem wise.

pkg/parser/grok/grok.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	g, err := grok.NewWithConfig(&grok.Config{NamedCapturesOnly: true})
+	if err != nil {
+		return nil, fmt.Errorf("new grok object: %w", err)
+	}
+
+	include := map[string]bool{}
+	exclude := map[string]bool{}
```
srikanthccv, line 68: `map[string]struct{}` is usually preferred over `map[string]bool{}` for membership testing scenarios since the former is memory efficient. It may not make a big difference here but it's good to follow that in general.

pkg/parser/grok/grok.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (r *Parser) match(value string) (interface{}, error) {
+	if r.cache != nil {
+		if x := r.cache.get(value); x != nil {
+			return x, nil
+		}
+	}
+
+	values, err := r.grok.Parse(r.pattern, value)
```
srikanthccv, line 138: This looks like all string values, is it considered when the user desires some parsed values to map to appropriate types such as status -> int etc.?

pkg/parser/grok/grok.go
nityanandagohain: Thanks for sharing, found this article https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b

pkg/parser/grok/cache.go
nityanandagohain: Well while building this parser, I took the same cache that was available for regex. But your question is legit.

Can create a new issue and we can have a discussion over there.

pkg/parser/grok/grok.go
nityanandagohain: Yeah as of now both the regex and this grok parser returns string values. For converting it to different datatype the attribute operator is used 
ex:- https://github.com/SigNoz/nginx-logs-parsing/blob/c14a2178a050293a50549b79991ed3858e295c78/clickhouse-setup/otel-collector-config.yaml#L123

srikanthccv: I saw some TypedParse or similar in the grok lib API so I was thinking If one would prefer that instead of adding another processor later.

nityanandagohain: Yeah, makes sense. Though if someone uses an existing pattern it might not contain corresponding types. But if they use patterns like number or string then they should be able to do it right here. Let me check and add it.

nityanandagohain: Added the support for this

pkg/parser/grok/cache.go
nityanandagohain: https://github.com/SigNoz/signoz-otel-collector/issues/71

srikanthccv (approved):

srikanthccv:
I think we should merge this. It doesn't need to wait for some users to ask about the capability.

PR #56, opened 2022-12-24
Add config manager

srikanthccv:

See first https://github.com/SigNoz/signoz-otel-collector/pull/51

This PR adds a config manager to the collector. The config manager is responsible for applying the remote configuration to the collector. It restarts the collector when the configuration is updated.

### DIFF
--- a/opamp/config_manager.go
+++ b/opamp/config_manager.go
@@ -1,9 +1,15 @@
 package opamp
 
 import (
+	"bytes"
+	"fmt"
+	"os"
+
 	"github.com/open-telemetry/opamp-go/protobufs"
 	"go.uber.org/zap"
 )
+
+const collectorConfigKey = "collector.yaml"
 
 // agentConfigManager is responsible for managing the agent configuration
 // It is responsible for:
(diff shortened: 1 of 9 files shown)


opamp/config_manager.go
```diff
@@ -38,5 +93,39 @@ func (a *agentConfigManager) CreateEffectiveConfigMsg() (*protobufs.EffectiveCon
...
+
+	if !ok {
+		return false, nil
+	}
+
+	return a.applyRemoteConfig(a.agentConfig, remoteCollectorConfig.GetBody())
+}
+
+// applyRemoteConfig applies the remote config to the agent.
+func (a *agentConfigManager) applyRemoteConfig(currentConfig *remoteControlledConfig, newContents []byte) (changed bool, err error) {
```
ankitnayan, line 112: Is the `currentConfig` as input param needed? Won't it be available as `a.agentConfig`?

ankitnayan:
```
type agentConfigManager struct {
	agentConfig *remoteControlledConfig
	logger      *zap.Logger
}
````
Shouldn't this be locked as there might be multiple concurrent events/calls that might want to change the agent config?

opamp/config_manager.go
srikanthccv: You are correct. This should have a simple function instead of a receiver method on the struct.

srikanthccv:
>Shouldn't this be locked as there might be multiple concurrent events/calls that might want to change the agent config?

It is enforced [here](https://github.com/open-telemetry/opamp-go/blob/main/server/types/callbacks.go#L17-L41) that for a single connection (i.e. one agent) the callbacks are never called concurrently so given that the lock is not used here. We can add that, but if the server is enforced to take care of that details, it's not necessary to synchronize here.

ankitnayan:
> > Shouldn't this be locked as there might be multiple concurrent events/calls that might want to change the agent config?
> 
> It is enforced [here](https://github.com/open-telemetry/opamp-go/blob/main/server/types/callbacks.go#L17-L41) that for a single connection (i.e. one agent) the callbacks are never called concurrently so given that the lock is not used here. We can add that, but if the server is enforced to take care of that details, it's not necessary to synchronize here.

What if multiple instances of query-service run? We must run opamp server as a separate service then? Adding lock here might be more robust. We can create a separate issue to handle this later if you want.

srikanthccv:
There are no locks anywhere in the client implementation. I think addressing it separately is a good idea. We will check if the safe concurrency guarantee is needed not just for the config manager but as a whole.

nityanandagohain (approved):
LGTM for now.

PR #57, opened 2022-12-24
Add server client implementation

srikanthccv:

See first https://github.com/SigNoz/signoz-otel-collector/pull/56

This PR adds "serverClient" implementation that connects to the Opamp server and manages the agent configuration.


opamp/server_client.go
```diff
@@ -25,33 +30,201 @@ type serverClient struct {
...
 // It stops the Opamp client and disconnects from the Opamp server
 func (s *serverClient) Stop(ctx context.Context) error {
-	return nil
+	s.collector.Shutdown()
+	return s.opampClient.Stop(ctx)
 }
 
 // onMessageFuncHandler is the callback function that is called when the Opamp client receives a message from the Opamp server
 func (s *serverClient) onMessageFuncHandler(ctx context.Context, msg *types.MessageData) {
+	if msg.RemoteConfig != nil {
```
ankitnayan, line 160: Should we handle more fields in the `types.MessageData`? https://github.com/open-telemetry/opamp-go/blob/main/client/types/callbacks.go#L9

srikanthccv: Yes, but our current focus is on the remote config, so we look for that in the received message. I left a TODO note below so the reader knows that there are more capabilities.

ankitnayan: that would be good

opamp/server_client.go
```diff
@@ -25,33 +30,210 @@ type serverClient struct {
...
+
+		remoteCfgStatus.Status = protobufs.RemoteConfigStatuses_RemoteConfigStatuses_FAILED
+		remoteCfgStatus.ErrorMessage = fmt.Sprintf("failed to apply config changes: %s", err.Error())
+	}
+
+	if err := s.opampClient.SetRemoteConfigStatus(remoteCfgStatus); err != nil {
+		return fmt.Errorf("failed to set remote config status: %w", err)
+	}
+
+	if changed {
```
mindhash, line 197: It would have been good if opamp supported sending effective config along with remote config status. 

In the query service, we use effective config for any updates. If this calls fails, we might not have the current effective config on the server. Hence there can be mismatch. Though condition is an edge case. 

In query service side, may be we can sync pull instead of waiting for a push. Not an immediate need though.

mindhash (approved):

nityanandagohain (approved):
Looks good to me for now with the understanding I got. We can add more things as required in newer versions.

Issue #65, opened 2022-12-26
Metricsexporter tests fail with panic


srikanthccv:
Example test run https://github.com/SigNoz/signoz-otel-collector/actions/runs/3731714139/jobs/6330251346. This should be fixed either by passing the mock or the original ClickHouse connection.

srikanthccv:
Not relevant

PR #67, opened 2022-12-26
Temporarily skip some metrics exporter tests

srikanthccv:
They should be enabled when there is a proper fix for https://github.com/SigNoz/signoz-otel-collector/issues/65.

On a side note, I would like to ask if it's generally recommended/practised approach that the connection should be established in the new object instantiation with `NewXX` or when the `Start` is actually called on the exporter; even outside this project, is it common?

### DIFF
--- a/exporter/clickhousemetricsexporter/exporter_test.go
+++ b/exporter/clickhousemetricsexporter/exporter_test.go
@@ -41,7 +41,8 @@ import (
 )
 
 // Test_NewPRWExporter checks that a new exporter instance with non-nil fields is initialized
-func Test_NewPRWExporter(t *testing.T) {
+// FIXME(srikanthccv): Enable the tests once this issue is fixed: https://github.com/SigNoz/signoz-otel-collector/issues/65
+func skip_Test_NewPRWExporter(t *testing.T) {
 	cfg := &Config{
 		ExporterSettings:   config.NewExporterSettings(component.NewID(typeStr)),
 		TimeoutSettings:    exporterhelper.TimeoutSettings{},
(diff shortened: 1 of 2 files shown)


srikanthccv:
>Since we have a start function, I think the connection should be established in the Start instead of NewXX

Yes, +1. Say there wasn't a `Start`, would it still make sense to not open a connection in `NewXX` or instead pass a conn as a part of `NewXXSettings`? I don't know if there is an agreed practice but In my limited experience that is what I observed. It also helps with testing because I can instead pass a mock conn instead of a real conn for unit tests.

nityanandagohain:
> instead pass a conn as a part of NewXXSettings

can you give some more idea about this, an example or link to something will be helpful.

srikanthccv:
I am taking the metrics exporter example since it's where the change introduced is causing issues. The `NewPrwExporter` is intended to return an exporter implementation. 

https://github.com/SigNoz/signoz-otel-collector/blob/90d02971ed39a2abfbd2a2c368be54bcda9a7e43/exporter/clickhousemetricsexporter/exporter.go#L65-L91


It would be modified to take the `conn` as part of the params instead of opening Connection inside the function at line 85. If that were the implementation, one could provide the mock conn, and it still runs without throwing the nil pointer error.

srikanthccv:
I would appreciate approval on this if there is nothing pending.

nityanandagohain:
@srikanthccv would you suggest making the changes instead of skipping it ?

srikanthccv:
That will be a non-trivial change, so I created a separate issue address https://github.com/SigNoz/signoz-otel-collector/issues/65. I don't have the mock connection impl ready. I have the incomplete useable mock implementation here https://github.com/srikanthccv/ClickHouse-go-mock. I will add the proposed changes when that is ready to use.

nityanandagohain (approved):

Issue #74, opened 2023-01-03
Add exporter latency metric instrumentation


srikanthccv:
- [x] #75
- [x] #76
- [x] #77

PR #78, opened 2023-01-03
Add DB latency metric instrumentation for clickhousemetricsexporter

### LINKED ISSUE #75: Add exporter latency metric instrumentation for clickhouemetricsexporter

srikanthccv:
Part of #74; Fixes #75

This adds the DB write latency instrumentation for the metrics exporter. This helps in understanding the writing performance of ClickHouse. The table `time_series_v2` table doesn't see much ingestion because we maintain a list of written series in memory and conditionally insert them. However, we still instrument it; the `samples_v2` will be helpful to know if the ClickHouse is the reason for the slow writes. 

Since a few tens of millis is not sensitive for us here, the buckets start from 100ms and go up 512 seconds exponentially because the default send timeout is 300s.


exporter/clickhousemetricsexporter/clickhouse.go
```diff
@@ -313,8 +314,14 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest) erro
 			}
 		}
 
-		return statement.Send()
-
+		start := time.Now()
+		err = statement.Send()
```
ankitnayan, line 318: @srikanthccv is `statement.Send()` a blocking call?

ankitnayan (approved):

exporter/clickhousemetricsexporter/clickhouse.go
srikanthccv: Yes

nityanandagohain:
@srikanthccv we have a usage.go file in each of the exporters, which has just the views and measures for the metrics that help in usage collection.

I am thinking if we can move the view to that file and rename that file to metrics.go, any suggestions? That can be a common place for all the custom metrics we are creating for the exporter.

srikanthccv:
I am fine with the idea of common place for all custom metrics together but moving these latencies under usage doesn't sound good to me.

srikanthccv:
@nityanandagohain, I think I misread your suggestion. If I understand it correctly, you suggest creating a `metrics.go` file in each exporter and maintaining all the measurements and views there but only registering them in Factory? That's a good suggestion. I will address this separately in different PR.

nityanandagohain:
Right @srikanthccv.

PR #79, opened 2023-01-06
Add DB latency metric instrumentation for clickhouselogsexporter

nityanandagohain:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/76

similar to https://github.com/SigNoz/signoz-otel-collector/pull/78 but this is for logs

### DIFF
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -171,7 +171,13 @@ func (e *clickhouseLogsExporter) pushLogsData(ctx context.Context, ld plog.Logs)
 	if err != nil {
 		return fmt.Errorf("StatementSend:%w", err)
 	}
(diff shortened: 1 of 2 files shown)


exporter/clickhouselogsexporter/exporter.go
```diff
@@ -164,10 +165,20 @@ func (e *clickhouseLogsExporter) pushLogsData(ctx context.Context, ld plog.Logs)
...
 	if err != nil {
 		return fmt.Errorf("StatementSend:%w", err)
 	}
+	stats.RecordWithTags(ctx,
+		[]tag.Mutator{
+			tag.Upsert(exporterKey, string(component.DataTypeLogs)),
+			tag.Upsert(tableKey, DISTRIBUTED_LOGS_TABLE),
+		},
+		writeLatencyMillis.M(int64(time.Since(dbWriteStart).Milliseconds())),
+	)
```
srikanthccv, line 180: This should be before `err` return block. Say it errd with timeout you would want to record because that spike indicates something unusual. Not recording that means losing the good data point.

nityanandagohain: Yeah, makes sense for this metric.

PR #81, opened 2023-01-09
Add signoz.collector.id to span metrics processor

srikanthccv:
Part of https://github.com/SigNoz/signoz/issues/2000

### DIFF
--- a/processor/signozspanmetricsprocessor/factory.go
+++ b/processor/signozspanmetricsprocessor/factory.go
@@ -17,6 +17,7 @@ package signozspanmetricsprocessor
 import (
 	"context"
 
+	"github.com/google/uuid"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config"
 	"go.opentelemetry.io/collector/consumer"
@@ -28,6 +29,8 @@ const (
 	typeStr = "signozspanmetrics"
 	// The stability level of the processor.
 	stability = component.StabilityLevelBeta
+
+	signozID = "signoz.collector.id"
 )
 
 // NewFactory creates a factory for the spanmetrics processor.
@@ -49,5 +52,8 @@ func createDefaultConfig() component.ProcessorConfig {
 }
 
 func createTracesProcessor(_ context.Context, params component.ProcessorCreateSettings, cfg component.ProcessorConfig, nextConsumer consumer.Traces) (component.TracesProcessor, error) {
-	return newProcessor(params.Logger, cfg, nextConsumer)
+	// TODO(srikanthccv): use the instanceID from params when it is added
+	instanceUUID, _ := uuid.NewRandom()
+	instanceID := instanceUUID.String()
+	return newProcessor(params.Logger, instanceID, cfg, nextConsumer)
 }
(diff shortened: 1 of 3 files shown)


processor/signozspanmetricsprocessor/factory.go
```diff
@@ -28,6 +29,8 @@ const (
 	typeStr = "signozspanmetrics"
 	// The stability level of the processor.
 	stability = component.StabilityLevelBeta
+
+	signozID = "signoz.collector.id"
```
ankitnayan, line 33: @srikanthccv will this produce a dotted named key or underscored in final labels? I think we have been following underscore naming in labels till now.

srikanthccv: This will be converted to _ in the metrics exporter

ankitnayan: `resource_signoz_collector_id` and `signoz_collector_id` have duplicate values? How is it happening?

srikanthccv: If I recall correctly, we added a `resource_` prefix for all the attributes to be included so that they could be used for filtering by resource attributes on the services page.

srikanthccv: I am not sure how this is being consumed in the front end. I can look up to see if this can be removed https://github.com/SigNoz/signoz-otel-collector/blob/b0f42d0f8369d275fd51e83c73d7b1b64d8b2fa9/processor/signozspanmetricsprocessor/processor.go#L863-L871

srikanthccv: Inherited from

https://github.com/SigNoz/opentelemetry-collector-contrib/blob/c28c2913122ee25c131d29689544d61c9387ed92/processor/signozspanmetricsprocessor/processor.go#L687-L696

srikanthccv: I looked into frontend implementation. It makes a request to query service to match with `resource_` prefix and then strips away the `resource_` part to make the key look like a trace dot separated trace attribute. This is done just for the span metrics. I don't think I have the complete context behind the reasoning. We could get rid of `signoz_collector_id` for APM metrics from as far as I see, but I think this means we should prefix all resource attributes with `resource_` in metrics exporter which hasn't been the case with `resource_to_telemetry_conversion`.

https://github.com/SigNoz/signoz/blob/b23d8da96c2c21141528af72ea216e6f93cdfd5a/frontend/src/container/MetricsApplication/ResourceAttributesFilter/utils.ts#L24-L34

https://github.com/SigNoz/signoz/blob/b23d8da96c2c21141528af72ea216e6f93cdfd5a/frontend/src/lib/resourceAttributes.ts#L6-L16

ankitnayan: yeah...you got almost all pieces right. We add prefix to resource attributes and not add a copy. Here I see the exiting `signoz_collector_id` remains to exist.

srikanthccv: I am not sure if we can get rid of this duplication because there could be mixed use of them by users. I think I will open an issue for this, what do you think?

ankitnayan: yeah .. okay.. we can discuss them in another issue

ankitnayan (approved):

PR #85, opened 2023-01-18
Add support for bytes body type

nityanandagohain:
Fixes https://github.com/SigNoz/signoz/issues/2047

also added default observed timestamp, for receivers live filelog the observedtimestamp is set but Kafka doesn't sent default observedtimestamp.


srikanthccv (approved):
FYI: https://github.com/open-telemetry/opentelemetry-specification/pull/3014

Since its the official recommendation from spec to use body for whole structured data we may want to handle map type and creates attributes directly if not already present.

PR #89, opened 2023-01-25
feat: set default TTL

makeavish:
30days for metrics, 7 days for traces and logs

### DIFF
--- a/exporter/clickhousemetricsexporter/clickhouse.go
+++ b/exporter/clickhousemetricsexporter/clickhouse.go
@@ -101,7 +101,7 @@ func NewClickHouse(params *ClickHouseParams) (base.Storage, error) {
 		ENGINE = MergeTree
 			PARTITION BY toDate(timestamp_ms / 1000)
 			ORDER BY (metric_name, fingerprint, timestamp_ms)
-			TTL toDateTime(timestamp_ms) + toIntervalSecond(2592000);`, database, SAMPLES_TABLE, CLUSTER))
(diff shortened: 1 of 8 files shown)


exporter/clickhousemetricsexporter/clickhouse.go
```diff
@@ -100,7 +100,8 @@ func NewClickHouse(params *ClickHouseParams) (base.Storage, error) {
 		)
 		ENGINE = MergeTree
 			PARTITION BY toDate(timestamp_ms / 1000)
-			ORDER BY (metric_name, fingerprint, timestamp_ms);`, database, SAMPLES_TABLE, CLUSTER))
+			ORDER BY (metric_name, fingerprint, timestamp_ms)
+			TTL toDateTime(timestamp_ms) + toIntervalSecond(2592000);`, database, SAMPLES_TABLE, CLUSTER))
```
srikanthccv, line 104: Does this work correcly? Since the timestamp is in epoch millis?

makeavish: Good catch, let me update it.

makeavish: Updated @srikanthccv

srikanthccv (approved):

nityanandagohain:
Any reason to choose `30days for metrics, 7 days for traces and logs` . Or is it just that we  are expecting more storage to be used by traces and logs?

makeavish:
@nityanandagohain People generally do need to look at metrics older than 7 days for comparison, and it also takes up very less storage than logs and traces. We wanted to set retention low so that updating TTL later won't be costly operation.
We have updated docs and install script to let user know about default retention.

PR #93, opened 2023-02-10
chore: optimize multi architecture build

prashant-shahi:
This PR optimizes multi architecture build flow by building artifacts outside of container. 
Also, updated the depended workflows along with including intermediate `.build` folder in .gitignore file.

Additionally, sets `/logsmigrations` as default log migration folder from go constant.

Signed-off-by: Prashant Shahi <prashant@signoz.io>


prashant-shahi:
> Please add some description to give an idea of what has changed

PR description updated

exporter/clickhouselogsexporter/factory.go
```diff
@@ -33,7 +33,7 @@ const (
 	archiveNamespace = "clickhouselogs-archive"
 	databaseName     = "signoz_logs"
 	tableName        = "logs"
-	migrationsFolder = "./migrations"
+	migrationsFolder = "/logsmigrations"
```
srikanthccv, line 36: Running binary directly instead of docker image will not work this change?

cmd/signozcollector/Dockerfile
```diff
@@ -1,34 +1,25 @@
...
-ARG USER_UID=10001
-
-USER ${USER_UID}
-
-COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
-COPY --from=build /out/signoz-collector /
-COPY --from=build /src/exporter/clickhousetracesexporter/migrations /migrations
-COPY --from=build /src/exporter/clickhouselogsexporter/migrations /logsmigrations
-
-ENV LOG_MIGRATIONS_FOLDER="/logsmigrations"
```
srikanthccv, line 30: You probably have to revert this change. Try running the binary (important because building new image each time is time consuming) and check if the migrations run for logs.

cmd/signozcollector/Dockerfile
```diff
@@ -1,34 +1,25 @@
...
+# expose OTLP ports for the collector
+EXPOSE 4317
+EXPOSE 4318
 
-# Final stage
-FROM alpine:3.15
-
-ARG USER_UID=10001
-
-USER ${USER_UID}
```
srikanthccv, line 23: I forgot why this existed (I think something related ClickHouse), how does this removal of this affect?

Makefile
```diff
@@ -31,7 +34,18 @@ test:
 
 .PHONY: build
 build:
-	GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o signoz-collector ./cmd/signozcollector
+	CC=${GOCC} OS111MODULE=on CGO_ENABLED=1 GOOS=${GOOS} GOARCH=${GOARCH} go build -o .build/${GOOS}-${GOARCH}/signoz-collector -ldflags "-linkmode external -extldflags '-static' -s -w ${LD_FLAGS}" ./cmd/signozcollector
```
srikanthccv, line 37: `-tags timetzdata` is missing

exporter/clickhouselogsexporter/factory.go
prashant-shahi: In case of running directly, one can set the environment variable:

`export LOG_MIGRATIONS_FOLDER="./logsmigrations"`

For most case that is running in Docker/K8s, this newly introduced changes should work fine.
Also, keeping it consistent with the traces migration constant.

Makefile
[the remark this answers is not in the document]
prashant-shahi: Removing it as redundant. Since, that value should be set as `on` by default i.e. use Go modules.

exporter/clickhouselogsexporter/factory.go
srikanthccv: It works fine in docker/k8s, but that isn't my original concern. Generally, programs have defaults, and env/cli args are added to support the customisation. The stand-alone binary should run with defaults without any custom configuration, but this change doesn't make it work.

cmd/signozcollector/Dockerfile
prashant-shahi: This is not needed now. This made sense when we were using `scratch` base image, but with `alpine` it is not needed.

Makefile
prashant-shahi: nice catch

exporter/clickhouselogsexporter/factory.go
prashant-shahi: yeah, true. It's been set to `/migrations` as default for traces.
That should be updated.

Makefile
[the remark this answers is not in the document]
prashant-shahi: updated

Makefile
```diff
@@ -31,7 +34,18 @@ test:
...
+
+.PHONY: amd64
+amd64:
+	make GOARCH=amd64 build
+
+.PHONY: arm64
+arm64:
+	make GOCC=aarch64-linux-gnu-gcc GOARCH=arm64 build
```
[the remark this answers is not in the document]
prashant-shahi: OR support? you mean across amd/arm? It works on both of those. However, the GOOS env would be set to `linux` for both while building.

cmd/signozcollector/Dockerfile
```diff
@@ -1,34 +1,27 @@
...
 
-ARG TARGETPLATFORM
+# define arguments and default values
+ARG TARGETOS=linux
+ARG TARGETARCH=amd64
 
-RUN export GOOS=$(echo ${TARGETPLATFORM} | cut -d / -f1) && \
-    export GOARCH=$(echo ${TARGETPLATFORM} | cut -d / -f2)
+# copy the binaries from the multi-stage build
+COPY .build/${TARGETOS}-${TARGETARCH}/signoz-collector /signoz-collector
```
srikanthccv, line 12: Don't you have to set `TARGETOS` and `TARGETARCH` while building docker images?

prashant-shahi: We do not have to set them. It is automatically set globally by buildkit.

Reference: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

prashant-shahi:
moving the PR branch from fork to main repo and creating another PR

PR #100, opened 2023-02-21
fix: attribute name corrected in logs database

### LINKED ISSUE #94: Attribute spelling wrong in table name of logs
change atrribute to attribute

nityanandagohain:
Fixes #94

### DIFF
--- a/exporter/clickhouselogsexporter/migrations/000005_attribute_rename.down.sql
+++ b/exporter/clickhouselogsexporter/migrations/000005_attribute_rename.down.sql
@@ -1,10 +1,10 @@
-RENAME TABLE if exists signoz_logs.logs_attribute_keys TO signoz_logs.logs_atrribute_keys;
+RENAME TABLE IF EXISTS signoz_logs.logs_attribute_keys TO signoz_logs.logs_atrribute_keys;
 
 
 -- Materialized view
-DROP table if exists attribute_keys_string_final_mv;
-DROP table if exists attribute_keys_int64_final_mv;
-DROP table if exists attribute_keys_float64_final_mv;
+DROP VIEW IF EXISTS attribute_keys_string_final_mv;
+DROP VIEW IF EXISTS attribute_keys_int64_final_mv;
+DROP VIEW IF EXISTS attribute_keys_float64_final_mv;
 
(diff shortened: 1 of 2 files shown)


exporter/clickhouselogsexporter/migrations/000005_attribute_rename.down.sql
```diff
@@ -0,0 +1,32 @@
+RENAME TABLE if exists signoz_logs.logs_attribute_keys TO signoz_logs.logs_atrribute_keys;
```
srikanthccv, line 1: This is confusing honestly. I was wondering why you are adding a spelling mistake until I realized this is a .down file.

exporter/clickhouselogsexporter/migrations/000005_attribute_rename.down.sql
```diff
@@ -0,0 +1,32 @@
...
+
+
+-- Materialized view
+DROP table if exists attribute_keys_string_final_mv;
+DROP table if exists attribute_keys_int64_final_mv;
+DROP table if exists attribute_keys_float64_final_mv;
```
srikanthccv, line 7: Shouldn't these be `DROP VIEW IF EXISTS` instead of `DROP TABLE IF EXISTS`?

exporter/clickhouselogsexporter/migrations/000005_attribute_rename.up.sql
```diff
@@ -0,0 +1,32 @@
+RENAME TABLE if exists signoz_logs.logs_atrribute_keys TO signoz_logs.logs_attribute_keys;
+
+
+-- Materialized view
+DROP table if exists atrribute_keys_string_final_mv;
+DROP table if exists atrribute_keys_int64_final_mv;
+DROP table if exists atrribute_keys_float64_final_mv;
```
srikanthccv, line 7: Same question as above

exporter/clickhouselogsexporter/migrations/000005_attribute_rename.down.sql
nityanandagohain: Yea 😅  don't have any other option

exporter/clickhouselogsexporter/migrations/000005_attribute_rename.down.sql
nityanandagohain: Interestingly both works, but let me update it.

PR #102, opened 2023-02-21
feat: added signoz tail sampler first build

mindhash:
This PR includes tail sampling processor for traces - supports:
- config for policy groups and hierarchy of policies
- priority based processing
- early exits when a rule matches

### DIFF
--- a/components/components.go
+++ b/components/components.go
@@ -110,6 +110,7 @@ import (
 	"github.com/SigNoz/signoz-otel-collector/exporter/clickhousetracesexporter"
 	_ "github.com/SigNoz/signoz-otel-collector/pkg/parser/grok"
 	"github.com/SigNoz/signoz-otel-collector/processor/signozspanmetricsprocessor"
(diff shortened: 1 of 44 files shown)


srikanthccv:
Please split this up. There is no way one can know what the additions are vs the upstream copy. You could send one PR that just copies the upstream processor code followed by custom changes made by us. It makes it easier for the reviewers and faster to review the small CL and get it merged.

mindhash:
@srikanthccv  The upstream code is used as reference. Some mechanics are taken from it but we should treat this as a whole unit it self. 

> There is no way one can know what the additions are vs the upstream copy

I think We shouldn't add upstream processor code and its deps.  Upstream tail sampler works with a different processor name ("tail_sampling") than this ("signoz_tail_sampling"). This is done to ensure customer can use tail sampler (vanilla from otel contrib) with our custom one. 
 
> You could send one PR that just copies the upstream processor code followed by custom changes made by us.

PR #103, opened 2023-02-23
feat: push tag attributes to a new table

nityanandagohain:
Fixes https://github.com/SigNoz/signoz/issues/2254

### DIFF
--- a/exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
+++ b/exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
@@ -1,13 +1,14 @@
 CREATE TABLE IF NOT EXISTS signoz_logs.tag_attributes ON CLUSTER cluster (
     timestamp DateTime CODEC(ZSTD(1)), 
     tagKey LowCardinality(String) CODEC(ZSTD(1)),
-    tagType LowCardinality(String) CODEC(ZSTD(1)),
-    tagDataType LowCardinality(String) CODEC(ZSTD(1)),
+    tagType Enum('tag', 'resource') CODEC(ZSTD(1)),
+    tagDataType Enum('string', 'bool', 'int64', 'float64') CODEC(ZSTD(1)),
     stringTagValue String CODEC(ZSTD(1)),
-    numberTagValue Nullable(Float64) CODEC(ZSTD(1))
+    int64TagValue Nullable(Int64) CODEC(ZSTD(1)),
+    float64TagValue Nullable(Float64) CODEC(ZSTD(1))
 ) ENGINE ReplacingMergeTree
-ORDER BY (tagKey, tagType, tagDataType, stringTagValue, numberTagValue)
-TTL toDateTime(timestamp) + INTERVAL 86400 SECOND DELETE
+ORDER BY (tagKey, tagType, tagDataType, stringTagValue, int64TagValue, float64TagValue)
+TTL toDateTime(timestamp) + INTERVAL 172800 SECOND DELETE
 SETTINGS ttl_only_drop_parts = 1, allow_nullable_key = 1;
 
 CREATE TABLE IF NOT EXISTS signoz_logs.distributed_tag_attributes ON CLUSTER cluster AS signoz_logs.tag_attributes
--- a/exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
+++ b/exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
@@ -12,4 +12,4 @@ TTL toDateTime(timestamp) + INTERVAL 172800 SECOND DELETE
 SETTINGS ttl_only_drop_parts = 1, allow_nullable_key = 1;
 
 CREATE TABLE IF NOT EXISTS signoz_logs.distributed_tag_attributes ON CLUSTER cluster AS signoz_logs.tag_attributes
-ENGINE = Distributed("cluster", "signoz_logs", tag_attributes, cityHash64(tagKey));
+ENGINE = Distributed("cluster", "signoz_logs", tag_attributes, cityHash64(rand()));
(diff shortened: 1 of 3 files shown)


exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    tagDataType LowCardinality(String) CODEC(ZSTD(1)),
+    stringTagValue String CODEC(ZSTD(1)),
+    numberTagValue Nullable(Float64) CODEC(ZSTD(1))
+) ENGINE ReplacingMergeTree
+ORDER BY (tagKey, tagType, tagDataType, stringTagValue, numberTagValue)
+TTL toDateTime(timestamp) + INTERVAL 86400 SECOND DELETE
+SETTINGS ttl_only_drop_parts = 1, allow_nullable_key = 1;
+
+CREATE TABLE IF NOT EXISTS signoz_logs.distributed_tag_attributes ON CLUSTER cluster AS signoz_logs.tag_attributes
+ENGINE = Distributed("cluster", "signoz_logs", tag_attributes, cityHash64(tagKey));
```
srikanthccv, line 14: I think distributed for this task is an overkill.

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS signoz_logs.tag_attributes ON CLUSTER cluster (
+    timestamp DateTime CODEC(ZSTD(1)), 
+    tagKey LowCardinality(String) CODEC(ZSTD(1)),
+    tagType LowCardinality(String) CODEC(ZSTD(1)),
+    tagDataType LowCardinality(String) CODEC(ZSTD(1)),
```
srikanthccv, line 5: Both these could be enum.

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    stringTagValue String CODEC(ZSTD(1)),
+    numberTagValue Nullable(Float64) CODEC(ZSTD(1))
```
srikanthccv, line 7: How are booleans handled?

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
```diff
@@ -0,0 +1,14 @@
...
+) ENGINE ReplacingMergeTree
+ORDER BY (tagKey, tagType, tagDataType, stringTagValue, numberTagValue)
+TTL toDateTime(timestamp) + INTERVAL 86400 SECOND DELETE
```
srikanthccv, line 10: 2 days.

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
nityanandagohain: Yeah right it can be overkill, but don't have a strong opinion on this. Do you see any concerns going with distributed?

There might be cases where due to cardinality this table size will be high for logs data but even then a single table should handle it, but keeping it distributed will should make it future proof.

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
nityanandagohain: makes sense

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
nityanandagohain: Since the boolean value is either true/false, the API returns those two values. We are avoiding creating a new column for that. i.e the entry will be present but there won't be any value column for that.

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS signoz_logs.tag_attributes ON CLUSTER cluster (
+    timestamp DateTime CODEC(ZSTD(1)), 
+    tagKey LowCardinality(String) CODEC(ZSTD(1)),
+    tagType Enum('tag', 'resource') CODEC(ZSTD(1)),
+    tagDataType Enum('string', 'bool', 'int64', 'float64') CODEC(ZSTD(1)),
+    stringTagValue String CODEC(ZSTD(1)),
+    int64TagValue Nullable(Int64) CODEC(ZSTD(1)),
+    float64TagValue Nullable(Float64) CODEC(ZSTD(1))
```
nityanandagohain, line 8: Have added int64 and float64 @makeavish 

cc @srikanthccv

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
makeavish: Yeah I think we can use distributed.

srikanthccv: I would suggest pinking a better sharding_key as the`tagKey` may cause uneven distribution since some keys are likely to appear in data more than others, and if they are concentrated on one server, it nullifies the benefit of the distributed.

nityanandagohain: Made the changes, please add in traces as well @makeavish

makeavish (approved):

srikanthccv (approved):
One question, LGTM otherwise.

exporter/clickhouselogsexporter/migrations/000006_tag_attributes.up.sql
```diff
@@ -0,0 +1,15 @@
...
+) ENGINE ReplacingMergeTree
+ORDER BY (tagKey, tagType, tagDataType, stringTagValue, int64TagValue, float64TagValue)
+TTL toDateTime(timestamp) + INTERVAL 172800 SECOND DELETE
+SETTINGS ttl_only_drop_parts = 1, allow_nullable_key = 1;
+
+CREATE TABLE IF NOT EXISTS signoz_logs.distributed_tag_attributes ON CLUSTER cluster AS signoz_logs.tag_attributes
+ENGINE = Distributed("cluster", "signoz_logs", tag_attributes, cityHash64(rand()));
```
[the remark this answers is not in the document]
nityanandagohain: Thanks, done :)

srikanthccv (approved):

PR #111, opened 2023-03-07
feat: added signoz tail sampler

mindhash:
This PR introduces SigNoz tail sampler for traces.


processor/signoztailsampler/evaluator.go
```diff
@@ -0,0 +1,141 @@
...
+		if decision != sampling.NoResult {
+			// found a result, exit
+			return decision, nil
+		}
+	}
+
+	// filterMatched captures when atleast one filter matches
+	var filterMatched bool
+
+	for _, fe := range de.filters {
```
srikanthccv, line 112: It says `policy_filters` is what activates the policy, but here it matches after the sub-policies? Say I have a service where I want to do 20% probabilistic on key `my-custom-key-1` and 80% on ``my-custom-key-2` then my policy would be
```
{
	Name:     "test-policy-1",
	Type:     PolicyGroup,
	Priority: 1,
	Root:     true,
	PolicyFilterCfg: PolicyFilterCfg{
		StringAttributeCfgs: []StringAttributeCfg{
			{Key: "service.name", Values: []string{"route"}},
		},
	},
	SubPolicies: []PolicyCfg{
		{
			Name:     "custom-condition-1",
			Type:     PolicyGroup,
			Priority: 1,
			ProbabilisticCfg: ProbabilisticCfg{
				SamplingPercentage: 20,
			}
			PolicyFilterCfg: PolicyFilterCfg{
				StringAttributeCfgs: []StringAttributeCfg{
					{Key: "my-custom-key-1", Values: []string{"some-value-1"}},
				},
			},
			{
				Name:     "custom-condition-2",
				Type:     PolicyGroup,
				Priority: 1,
				ProbabilisticCfg: ProbabilisticCfg{
					SamplingPercentage: 80,
				}
				PolicyFilterCfg: PolicyFilterCfg{
					StringAttributeCfgs: []StringAttributeCfg{
						{Key: "my-custom-key-2", Values: []string{"some-value-2"}},
					},
				},
			},
		},
	},
},
```

Why does the evaluator go to sub-policies without making sure my service matches? Or did I misunderstand something? I don

processor/signoztailsampler/evaluator.go
```diff
@@ -0,0 +1,141 @@
...
+
+		if filterResult == sampling.Sampled {
+			filterMatched = true
+		}
+
+		if filterResult == sampling.NotSampled {
+			if de.filterOperator == "AND" {
+				// filter condition failed, return no op
+				return sampling.NoResult, nil
+			}
```
srikanthccv, line 130: Why specifically `AND`? I don't how `OR` will have a different result.

processor/signoztailsampler/factory.go
```diff
@@ -12,47 +12,39 @@
...
 		typeStr,
 		createDefaultConfig,
-		component.WithTracesProcessor(createTracesProcessor, stability))
+		component.WithTracesProcessor(createTracesProcessor, stability),
+	)
 }
 
 func createDefaultConfig() component.ProcessorConfig {
 	return &Config{
-		ProcessorSettings: config.NewProcessorSettings(component.NewID(typeStr)),
```
srikanthccv, line 53: Why is this removed?

processor/signoztailsampler/factory_test.go
```diff
@@ -27,11 +13,16 @@ import (
...
+// ToDo(amol):  commented this from otel-contrib/tail sampler
+// it results in stack overflow. this is maily because policy group
+// has self reference in form of subpolicies.
+// hence componenttest.CheckConfigStruct fails to validate config
+// rather keeps going in cycles to check the struct.
+// func TestCreateDefaultConfig(t *testing.T) {
+// 	cfg := createDefaultConfig()
+// 	assert.NotNil(t, cfg, "failed to create default config")
+// 	assert.NoError(t, componenttest.CheckConfigStruct(cfg))
+// }
```
srikanthccv, line 25: This shouldn't be TODO for later.

processor/signoztailsampler/internal/sampling/always_unsample.go
```diff
@@ -0,0 +1,25 @@
+package sampling
+
+import (
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.uber.org/zap"
+)
+
+type alwaysUnSample struct {
```
srikanthccv, line 8: nit:
```suggestion
type neverSample struct {
```

processor/signoztailsampler/internal/sampling/always_unsample.go
```diff
@@ -0,0 +1,25 @@
...
+// NewAlwaysSample creates a policy evaluator the samples all traces.
+func NewAlwaysUnsample(logger *zap.Logger) PolicyEvaluator {
+	return &alwaysSample{
+		logger: logger,
+	}
+}
+
+// Evaluate looks at the trace data and returns a corresponding SamplingDecision.
+func (aus *alwaysUnSample) Evaluate(pcommon.TraceID, *TraceData) (Decision, error) {
+	aus.logger.Debug("Evaluating spans in always-sample filter")
```
srikanthccv, line 23:
```suggestion
	aus.logger.Debug("Evaluating spans in never-sample filter")
```

processor/signoztailsampler/metrics.go
```diff
@@ -44,91 +41,91 @@ var (
...
-	}
-
-	sampledTagKeys := []tag.Key{tagPolicyKey, tagSampledKey}
-	countTracesSampledView := &view.View{
-		Name:        obsreport.BuildProcessorCustomMetricName(typeStr, statCountTracesSampled.Name()),
-		Measure:     statCountTracesSampled,
-		Description: statCountTracesSampled.Description(),
-		TagKeys:     sampledTagKeys,
-		Aggregation: view.Sum(),
-	}
```
srikanthccv, line 98: Why is this removed, but the `statCountTracesSampled` measure is used?

processor/signoztailsampler/evaluator.go
mindhash: In case of OR, if filterResult is set even once, the outcome will be TRUE. 
in case of AND, we dont want even a single negative result.

processor/signoztailsampler/evaluator.go
```diff
@@ -0,0 +1,141 @@
...
+	filterOperator := policyCfg.PolicyFilterCfg.FilterOp
+
+	// list of filter evaluators used to decide if this policy
+	// should be applied
+	filters := prepareFilterEvaluators(logger, policyCfg.PolicyFilterCfg)
+	// todo(amol): need to handle situations with zero filters
+
+	// list of sub-policies evaluators
+	subpolicies := make([]sampling.PolicyEvaluator, len(policyCfg.SubPolicies))
+	for _, subRule := range policyCfg.SubPolicies {
```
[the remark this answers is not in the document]
mindhash: At the moment, we only intent to support one level of sub-policies.   

The structure of policies and sub-policies is similar and the methods of evaluation too. Hence there is a self reference.

processor/signoztailsampler/evaluator.go
```diff
@@ -0,0 +1,141 @@
...
+func prepareFilterEvaluators(logger *zap.Logger, policyFilterCfg PolicyFilterCfg) []sampling.PolicyEvaluator {
+	var filterEvaluators []sampling.PolicyEvaluator
+	for _, s := range policyFilterCfg.StringAttributeCfgs {
+		filterEvaluators = append(filterEvaluators, sampling.NewStringAttributeFilter(logger, s.Key, s.Values, s.EnabledRegexMatching, s.CacheMaxSize, s.InvertMatch))
+	}
+
+	for _, n := range policyFilterCfg.NumericAttributeCfgs {
+		filterEvaluators = append(filterEvaluators, sampling.NewNumericAttributeFilter(logger, n.Key, n.MinValue, n.MaxValue))
+	}
+	// todo: status filter
```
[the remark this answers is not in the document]
mindhash: I will add a separate issue

processor/signoztailsampler/evaluator.go
```diff
@@ -0,0 +1,141 @@
...
+	samplingMethod string
+
+	filterOperator string
+	filters        []sampling.PolicyEvaluator
+
+	// the subpolicy evaluators will be processed in order of priority.
+	// the sub-policies will be evaluated prior to evaluating top-level policy,
+	// if any of subpolicy filters match, the sampling method associated with that
+	// sub-policy will be applied. and no further processing will be performed.
+	subpolicies []sampling.PolicyEvaluator
```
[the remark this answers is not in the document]
mindhash: this needs correction. The root filter is not being evaluated up-front. I will fix this.

processor/signoztailsampler/config_test.go
```diff
@@ -42,136 +27,34 @@ func TestLoadConfig(t *testing.T) {
...
-								sharedPolicyCfg: sharedPolicyCfg{
-									Name:               "test-and-policy-2",
-									Type:               StringAttribute,
-									StringAttributeCfg: StringAttributeCfg{Key: "key2", Values: []string{"value1", "value2"}},
-								},
-							},
+					PolicyFilterCfg: PolicyFilterCfg{
+						StringAttributeCfgs: []StringAttributeCfg{
+							{Key: "source", Values: []string{"security"}},
 						},
```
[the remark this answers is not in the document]
mindhash: yes. If sub-policy has a sampling rate different then it will take precedence. If none of the sub-policies apply then the root policy sampling rate will be applied.

processor/signoztailsampler/README.md
```diff
@@ -37,11 +35,20 @@ The following configuration options can also be modified:
 - `num_traces` (default = 50000): Number of traces kept in memory
 - `expected_new_traces_per_sec` (default = 0): Expected number of new traces (helps in allocating data structures)
```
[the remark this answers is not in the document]
mindhash: Ok. will do

processor/signoztailsampler/config.go
```diff
@@ -48,45 +48,10 @@ const (
 	SpanCount PolicyType = "span_count"
```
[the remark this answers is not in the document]
mindhash: ok

mindhash:
@srikanthccv I have identified these changes, will be working on them. 

- Add Config.Validate() to verify config 
- Add Root Filter prior to subpolicy filter with test case
- Look into CreateDefaultConfig test 
- update readme.md

processor/signoztailsampler/evaluator.go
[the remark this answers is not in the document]
srikanthccv: It would be good to maintain a separate struct for them; the README says policies will be evaluated in a depth-first manner, and this config allows any level of depth and code is recursing. I would suggest defining a sub-policy struct without self-referencing sub policy. This can also become the regular user's processor, not just the sampling rules from UI.

processor/signoztailsampler/evaluator_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+	require.NoError(t, err1, "failed to evaluate trace for policy with zero sampling")
+	require.Equal(t, sampling.NotSampled, decision1, "expected to not sampled as sampling percent is zero")
+
+	cfg.SamplingPercentage = 99
+	evaluator99 := NewDefaultEvaluator(zap.NewNop(), &cfg)
+	decision99, err99 := eval(evaluator99, traceIds[0], traces[0])
+
+	require.NoError(t, err99, "failed to evaluate trace for policy with 99 percent sampling")
+	require.Equal(t, sampling.Sampled, decision99, "expected to sampled as sampling percent is 99, re-run the test to try again")
```
srikanthccv, line 118: Please max of X retries here itself to reduce the flakiness of tests.

processor/signoztailsampler/config.go
```diff
@@ -194,9 +136,40 @@ type SpanCountCfg struct {
...
+
+	Priority int `mapstructure:"priority"`
+
+	// sampling applied when  PolicyFilter matches
+	ProbabilisticCfg `mapstructure:"sampling"`
+
+	// filter to activate policy
+	PolicyFilterCfg `mapstructure:"policy_filter"`
+
+	SubPolicies []PolicyCfg `mapstructure:"sub_policies"`
```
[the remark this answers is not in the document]
mindhash: The sub-policies follow the same semantics of root policy. 

The approach is derived from prom alert manager routes and it was picked to support rule hierarchy. 
https://prometheus.io/docs/alerting/latest/configuration/#route

I don't see a reason to limit this feature to one level through architecture. It can be a choice at higher level - through config or other means. 

If understanding is issue, I will document the code in further detail.

processor/signoztailsampler/README.md
```diff
@@ -160,25 +167,25 @@ processors:
...
-You are already incurring the cost of running the tail sampling processor, adding the probabilistic policy will be negligible.
-Additionally, using the policy within the tail sampling processor will ensure traces that are sampled by other policies will not be dropped.
+...you are already using the tail sampling processor: add the probabilistic sampling policy. You are already incurring the cost of running the tail sampling processor, adding the probabilistic policy will be negligible. Additionally, using the policy within the tail sampling processor will ensure traces that are sampled by other policies will not be dropped.
 
 [beta]: https://github.com/open-telemetry/opentelemetry-collector#beta
 [contrib]: https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib
-[probabilistic_sampling_processor]: ../probabilisticsamplerprocessor
\ No newline at end of file
+[probabilistic_sampling_processor]: ../probabilisticsamplerprocessor
+[loadbalancing_exporter]: ../../exporter/loadbalancingexporter
```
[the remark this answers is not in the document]
mindhash: Changed the readMe now. This link wont appear.

processor/signoztailsampler/internal/sampling/always_unsample.go
mindhash: Not sure I want to change the name and spend time on testing this. I can schedule this as time permits. Not committing the change for now.

mindhash: this is done.

processor/signoztailsampler/factory_test.go
mindhash: Handled this now. Removed the self-ref from policy and this test goes through ok.

processor/signoztailsampler/evaluator.go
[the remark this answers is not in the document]
mindhash: this is now handled. the self-ref is removed.

processor/signoztailsampler/evaluator.go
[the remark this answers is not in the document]
mindhash: this is now fixed. the policy filter gets evaluated upfront before sub-policies are looked into

processor/signoztailsampler/README.md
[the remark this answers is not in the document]
mindhash: removed un-used policy types.

processor/signoztailsampler/config.go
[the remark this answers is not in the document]
mindhash: removed.

processor/signoztailsampler/config.go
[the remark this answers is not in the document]
mindhash: this is done. removed self-ref in policy

processor/signoztailsampler/evaluator_test.go
mindhash: I am not sure how to handle this. Please do suggest the approach in detail. 
I can increase the sampling rate to 99.99 or re-try a few number of times to see majority of them succeed.

processor/signoztailsampler/config.go
```diff
@@ -209,4 +182,7 @@ type Config struct {
 	// PolicyCfgs sets the tail-based sampling policy which makes a sampling decision
 	// for a given trace when requested.
 	PolicyCfgs []PolicyCfg `mapstructure:"policies"`
+
+	// read only version number (optional)
+	Version int
 }
```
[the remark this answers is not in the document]
mindhash: this is done. At the moment, the validation is simple, We just expect priority to be explicitly assigned and should be > 1. 

I don't want to add too many validations without seeing how the UI evolves.

processor/signoztailsampler/metrics.go
srikanthccv: The opencensus is already being used by exporters. It shouldn't add any unnecessary dependencies? And This view is necessary because you are using the corresponding counter measure `statCountTracesSampled` in the processor.go. We need to keep the views for all the measures in processor.

processor/signoztailsampler/testdata/tail_sampling_config.yaml
```diff
@@ -1,107 +1,30 @@
...
+          policy_filter: {
+            string_attributes: {
+              key: source,
+              values: ["security"]
+            }
+          }, 
+          sampling: {sampling_percentage: 100},
+          sub_policies: [{
+            name: condition-1, 
+            type: policy_group,
```
srikanthccv, line 20: Can you add additional config tests with different policy types? I think we only have three or four.

srikanthccv (commented):
Thanks for addressing my comments. This overall LGTM. Left some comments.

srikanthccv:
@mindhash Please take a look at the unresolved conversations. Apart from some minor suggestions, there are two questions about checking the `PolicyType` for the sampler and OpenCensus view.

processor/signoztailsampler/config.go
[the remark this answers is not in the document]
mindhash: done

processor/signoztailsampler/metrics.go
mindhash: I added the view back. Not sure how to test this but it is now included.

processor/signoztailsampler/factory.go
mindhash: added back

processor/signoztailsampler/evaluator.go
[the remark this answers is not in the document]
mindhash: removed.

mindhash:
@srikanthccv I addressed all the comments that I could.

processor/signoztailsampler/metrics.go
srikanthccv: There is a counter measurement `statCountTracesSampled` here that is tracking something, and this is the view for it. It should be registered for the collector to emit the metric data. And to test it, If you run the collector and use this processor, you should see the metric in the `samples_v2` table of ClickHouse.

processor/signoztailsampler/factory.go
srikanthccv: You probably haven't pushed the changes, please do.

processor/signoztailsampler/evaluator.go
```diff
@@ -0,0 +1,162 @@
...
+		subEvaluators = append(subEvaluators, subEvaluator)
+	}
+
+	// sampling is applied only when filter conditions are met
+	var sampler sampling.PolicyEvaluator
+	var samplingMethod string
+
+	// todo(amol): need to handle cases where percent is not set by
+	// the user
+	switch policyCfg.SamplingPercentage {
```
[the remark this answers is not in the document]
srikanthccv: Here is the config.go list of `PolicyType`s https://github.com/SigNoz/signoz-otel-collector/pull/111/files#diff-090d391f2e2601a4d33c28f453a6d86e3d1105a5303db31b9cd9c72e12a33df2R28-R43

They are 
- always_sample
- never_sample
- numeric_attribute
- probabilistic
- string_attribute


There are evaluators such as  `sampling.NewStringAttributeFilter` and `sampling.NewNumericAttributeFilter` so I imagine this list of policy types is correct. And I would expect that the `SamplingPercentage` to be effective only when the select policy type is `probabilistic`

processor/signoztailsampler/testdata/tail_sampling_config.yaml
srikanthccv: See my comment here https://github.com/SigNoz/signoz-otel-collector/pull/111/files#r1146944199. In the config.go, there are five policy types defined, and any config is valid as long it belongs to one of them. If you are saying that there should only be `policy_group`, you should remove the rest.

processor/signoztailsampler/config.go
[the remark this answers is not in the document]
srikanthccv: Thanks.

srikanthccv:
Thank you, I still have questions about the policy types. Please take a look at the comments.

srikanthccv (approved):

cmd/signozcollector/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Builder stage
-FROM golang:1.19-alpine as build
+FROM golang:1.19 as build
```
prashant-shahi, line 2: @srikanthccv can you revert this particular change?

`push-workflow` is failing because the rest of the steps in Dockerfile assumes the base image to be alpine.

prashant-shahi (commented):
left comments regarding Dockerfile.

Also, I was wondering about deletion of `and_helper.go` and `composite_helper.go` files in `signoztailsampler`. 
Were they intentional?

cmd/signozcollector/Dockerfile
prashant-shahi: Anyways, this should be resolved when we merge PR #136

srikanthccv:
> Were they intentional?

Yes, the way signoz tail sampler works and the upstream tail sampler works are very different.

PR #114, opened 2023-03-09
fix: tests do not require db

srikanthccv:
This change removes the usage of `clickhouse{traces,metricswrite}` exporters in tests. Otherwise, tests fail in when the connection is not available. There is nothing about the ClickHouse in these tests, so there is no point in starting up a container for DB during the CI run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@
 .DS_Store
 .idea
 ./config/default-config.yaml
+*.rollback
+signoz-collector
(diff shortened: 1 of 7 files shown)


srikanthccv:
Please take a look; the failing tests in the PR will be fixed in https://github.com/SigNoz/signoz-otel-collector/pull/111.

nityanandagohain (approved):

PR #118, opened 2023-03-22
Add new resource attributes map column

srikanthccv:
- Add `resourceTagsMap` to `signoz_index_v2`
   - Since the `dependency_graph_minutes` is mv on `signoz_index_v2`, we also needed to add the resource attributes column to that table. 

- Need to change the primary key `dependency_graph_minutes` (can't, so we create a new table)
   - If we want to add a resource attributes filtering on the service map, we must introduce the resource attributes column to mv. The primary key, which is used to maintain the aggregate, will also change as we want to have a separate state for different resource attributes (i.e. we don't want to merge the dev and staging; otherwise, the user won't be able to see the data for one env, this allows them to see the individual env or all envs combined although they don't usually look at all envs combined). 

- Create a new table `dependency_graph_minutes_v2`.
  - The `dependency_graph_minutes_v2` is an `AggregatingMergeTree` that contains the temporary aggregate state for (minute, src, dest, src_resource_hash, dest_resource_hash). 
  - Why are source and destination attributes and their hash required? 
  - When you filter by attributes such as cluster name or region or similar, we only want to limit the result for that cluster/region. Showing for all regions would be misleading.

A little unsure about the perf implications because this mv involves storing the resource attributes and their hashes generated on the fly.

### DIFF
--- a/exporter/clickhousetracesexporter/migrations/000022_dependency_graph_minutes_v2.up.sql
+++ b/exporter/clickhousetracesexporter/migrations/000022_dependency_graph_minutes_v2.up.sql
@@ -5,17 +5,12 @@ CREATE TABLE IF NOT EXISTS signoz_traces.dependency_graph_minutes_v2 ON CLUSTER 
     error_count SimpleAggregateFunction(sum, UInt64) CODEC(T64, ZSTD(1)),
     total_count SimpleAggregateFunction(sum, UInt64) CODEC(T64, ZSTD(1)),
     timestamp DateTime CODEC(DoubleDelta, LZ4),
-    src_resource_attributes Map(LowCardinality(String), String) CODEC(ZSTD(1)),
-    src_resource_attributes_hash UInt64 CODEC(Delta, ZSTD(1)),
-    dest_resource_attributes Map(LowCardinality(String), String) CODEC(ZSTD(1)),
-    dest_resource_attributes_hash UInt64 CODEC(Delta, ZSTD(1)),
-    INDEX idx_src_resource_attributes_keys mapKeys(src_resource_attributes) TYPE bloom_filter(0.01) GRANULARITY 64,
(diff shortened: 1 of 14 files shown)


srikanthccv:
Did you mean the `order by`? How do you think that impacts negatively?

exporter/clickhousetracesexporter/migrations/000022_dependency_graph_minutes_v2.up.sql
```diff
@@ -0,0 +1,74 @@
...
+    countIf(B.statusCode=2) as error_count,
+    count(*) as total_count,
+    toStartOfMinute(B.timestamp) as timestamp,
+    anyLast(A.resourceTagsMap) as src_resource_attributes,
+    farmFingerprint64(A.resourceTagsMap) as src_resource_attributes_hash,
+    anyLast(B.resourceTagsMap) as dest_resource_attributes,
+    farmFingerprint64(B.resourceTagsMap) as dest_resource_attributes_hash
+FROM signoz_traces.signoz_index_v2 AS A, signoz_traces.signoz_index_v2 AS B
+WHERE (A.serviceName != B.serviceName) AND (A.spanID = B.parentSpanID)
+GROUP BY timestamp, src, dest, src_resource_attributes_hash, dest_resource_attributes_hash;
```
ankitnayan, line 37: @srikanthccv I meant this `group by` clause

srikanthccv:
Update based on our conversation, PTAL.

makeavish (approved):

makeavish:
@srikanthccv Can we drop `distributed_dependency_graph_minutes` table and MVs as they are not used now?

srikanthccv:
Yes, we could do that. I was thinking of deprecating it in the release notes and drop in the next release after the resource attributes release, just in case someone used that to write custom queries.

srikanthccv:
We will give a grace time before dropping the table.

PR #124, opened 2023-04-02
Add logs DB creation/deletion in migrations

RealHarshThakur:
Having schema migrations managed outside OTel exporter has its advantages:
* OTel exporter can have RBAC such that  its restricted from creation of databases, tables, etc- to follow the priciple of least privilege
* Other Schema versioning tools could be used to avoid schema conflicts. Here's a detailed explanation of the problem : https://atlasgo.io/concepts/migration-directory-integrity . 
I'm looking at adding support for Clickhouse in Atlas so we can manage it efficiently and automate the process. 

This MR does the following:
* Adds logs database creation/deletion in migration directory
* Adds an environment variable to explicitly opt-out of collector applying migrations. This maintains backward compatibility so that existing setups aren't impacted.

### DIFF
--- a/exporter/clickhouselogsexporter/migrations/000001_init_db.down.sql
+++ b/exporter/clickhouselogsexporter/migrations/000001_init_db.down.sql
@@ -5,3 +5,4 @@ DROP TABLE IF EXISTS signoz_logs.resource_keys_string_final_mv ON CLUSTER cluste
 DROP TABLE IF EXISTS signoz_logs.atrribute_keys_float64_final_mv ON CLUSTER cluster;
 DROP TABLE IF EXISTS signoz_logs.atrribute_keys_int64_final_mv ON CLUSTER cluster;
 DROP TABLE IF EXISTS signoz_logs.atrribute_keys_string_final_mv ON CLUSTER cluster;
+DROP DATABASE signoz_logs ON CLUSTER cluster;
--- a/exporter/clickhouselogsexporter/README.md
+++ b/exporter/clickhouselogsexporter/README.md
@@ -28,6 +28,10 @@ The following settings can be optionally configured:
     - `max_interval` (default = 30s): The upper bound on backoff; ignored if `enabled` is `false`
     - `max_elapsed_time` (default = 300s): The maximum amount of time spent trying to send a batch; ignored if `enabled` is `false`
 
+### Environment variable configuration options
+* LOG_MIGRATIONS_FOLDER: folder containing migrations to be applied by the OTel exporter
+* LOG_MIGRATIONS: when set to "false", the migrations will not be applied by OTel exporter
+
 ## Example
 
 ```yaml
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -330,6 +330,11 @@ func newClickhouseClient(logger *zap.Logger, cfg *Config) (clickhouse.Conn, erro
 		return nil, err
 	}
 
+	// in cases where we want migrations to be handled outside the OTel collector(for RBAC, schema versioning, etc)
+	if strings.ToLower(os.Getenv("LOGS_MIGRATE")) == "false" {
+		return db, nil
+	}
+
 	q := fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s ON CLUSTER %s;", databaseName, CLUSTER)
 	err = db.Exec(ctx, q)
 	if err != nil {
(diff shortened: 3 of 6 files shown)


exporter/clickhouselogsexporter/migrations/000001_init_db.down.sql
```diff
@@ -4,4 +4,5 @@ DROP TABLE IF EXISTS signoz_logs.logs_resource_keys ON CLUSTER cluster;
 DROP TABLE IF EXISTS signoz_logs.resource_keys_string_final_mv ON CLUSTER cluster;
 DROP TABLE IF EXISTS signoz_logs.atrribute_keys_float64_final_mv ON CLUSTER cluster;
 DROP TABLE IF EXISTS signoz_logs.atrribute_keys_int64_final_mv ON CLUSTER cluster;
-DROP TABLE IF EXISTS signoz_logs.atrribute_keys_string_final_mv ON CLUSTER cluster;
\ No newline at end of file
+DROP TABLE IF EXISTS signoz_logs.atrribute_keys_string_final_mv ON CLUSTER cluster;
+DROP DATABASE signoz_logs ON CLUSTER cluster;
```
srikanthccv, line 8: We don't change the existing migrations. It changes the hash and leaves the state to be dirty?

srikanthccv (commented):
>Here's a detailed explanation of the problem : https://atlasgo.io/concepts/migration-directory-integrity .

This is not really a problem for us in the foreseeable future. 

>Having schema migrations managed outside OTel exporter has its advantages

>OTel exporter can have RBAC such that its restricted from creation of databases, tables, etc

RBAC is not something we have made a priority, although we talked about it at some point, mainly because we ship with ClickHouse in our deployment and take it for granted. We do certain tasks outside the collector, which require us to make updates/delete operations on the table, and such levels of privileges are required to make SigNoz work. So if this is something users are concerned about and want to see it addressed, I would think about the broader effects and changes needed. There are also migrations for traces, and there is no migrations directory for metrics for legacy reasons.

exporter/clickhouselogsexporter/migrations/000001_init_db.down.sql
RealHarshThakur: > I would think about the broader effects and changes needed. There are also migrations for traces, and there is no migrations directory for metrics for legacy reasons.

It's definitely a broader issue to tackle- hopefully, we can do one at a time and make sure each change is backwards-compatible. I have looked only at the logs exporter and looks like it only needs these permissions for the creation of DB, tables during startup.

> We don't change the existing migrations. It changes the hash and leaves the state to be dirty?

Oh, my bad. I've only tried on a clean installation. So to remain backward-compatible, would creating a dedicated directory work?

srikanthccv: >It's definitely a broader issue to tackle- hopefully, we can do one at a time and make sure each change is backwards-compatible.

For such broader issues, it's better to have a plan and list the tasks before starting to work on anything. This will give us a chance to think about and evaluate it. Whether or not we make it backwards-compatible or to what extent it will be backwards-compatible, we can take a call but I believe this should be discussed and evaluated before we implement a solution.

>So to remain backward-compatible, would creating a dedicated directory work?

No, a new directory doesn't change anything. The version number and content of each migration file are important. The version can be any unsigned integer. I do not see a version `0` so we could move the creation of the database to something like `000000_create_db.{up/down}.sql`.

RealHarshThakur: Absolutely, what's the best way for me to be involved in discussing this? Are there Signoz community calls, etc?

srikanthccv: Can you create an issue over here https://github.com/SigNoz/signoz. As I understand, the main ask is that RBAC-controlled ClickHouse should work with SigNoz. Along with the migrations in the collector, we have DDL manipulation in the main repo for enabling the TTL setting from the UI.

RealHarshThakur: Done! https://github.com/SigNoz/signoz/issues/2556

RealHarshThakur: >  I do not see a version 0 so we could move the creation of the database to something like 000000_create_db.{up/down}.sql.

Did this, and it works now on existing and clean installs. :D

nityanandagohain:
Closing this as we have moved migrations out of the exporters https://github.com/SigNoz/signoz-otel-collector/tree/main/migrationmanager

PR #126, opened 2023-04-14
introduce span attributes keys table

makeavish:
This table will be used for cases when user writes a custom key

### DIFF
--- a/exporter/clickhousetracesexporter/migrations/000023_span_attributes_mv.up.sql
+++ b/exporter/clickhousetracesexporter/migrations/000023_span_attributes_mv.up.sql
@@ -7,7 +7,7 @@ CREATE TABLE IF NOT EXISTS signoz_traces.span_attributes ON CLUSTER cluster (
     float64TagValue Nullable(Float64) CODEC(ZSTD(1)),
     isColumn bool CODEC(ZSTD(1)),
 ) ENGINE ReplacingMergeTree
-ORDER BY (tagKey, tagType, dataType, stringTagValue, float64TagValue)
+ORDER BY (tagKey, tagType, dataType, stringTagValue, float64TagValue, isColumn)
 TTL toDateTime(timestamp) + INTERVAL 172800 SECOND DELETE
 SETTINGS ttl_only_drop_parts = 1, allow_nullable_key = 1;
 
--- a/exporter/clickhousetracesexporter/migrations/000024_span_attributes_keys.up.sql
+++ b/exporter/clickhousetracesexporter/migrations/000024_span_attributes_keys.up.sql
@@ -1,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS signoz_traces.span_attributes_keys ON CLUSTER cluster (
+    tagKey LowCardinality(String) CODEC(ZSTD(1)),
+    tagType Enum('tag', 'resource') CODEC(ZSTD(1)),
+    dataType Enum('string', 'bool', 'float64') CODEC(ZSTD(1)),
+    isColumn bool CODEC(ZSTD(1)),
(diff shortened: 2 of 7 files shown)


exporter/clickhousetracesexporter/migrations/000023_span_attributes_mv.up.sql
```diff
@@ -7,7 +7,7 @@ CREATE TABLE IF NOT EXISTS signoz_traces.span_attributes ON CLUSTER cluster (
     float64TagValue Nullable(Float64) CODEC(ZSTD(1)),
     isColumn bool CODEC(ZSTD(1)),
 ) ENGINE ReplacingMergeTree
-ORDER BY (tagKey, tagType, dataType, stringTagValue, float64TagValue)
+ORDER BY (tagKey, tagType, dataType, stringTagValue, float64TagValue, isColumn)
```
srikanthccv, line 10: Why is this needed here?

exporter/clickhousetracesexporter/writer.go
```diff
@@ -167,6 +170,13 @@ func (w *SpanWriter) writeBatch(batch []*Span) error {
 			return err
 		}
 	}
+	if w.attributeKeyTable != "" {
```
srikanthccv, line 173: The serial insertions, one at a time, will affect the overall throughput. It wouldn't have been an issue if the batch consisted only attributes because they are relatively small, but here it is iterating on the whole span batch and preparing the insertion.

exporter/clickhousetracesexporter/migrations/000024_span_attributes_keys.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS signoz_traces.span_attributes_keys ON CLUSTER cluster (
+    tagKey LowCardinality(String) CODEC(ZSTD(1)),
+    tagType Enum('tag', 'resource') CODEC(ZSTD(1)),
+    dataType Enum('string', 'bool', 'float64') CODEC(ZSTD(1)),
+    isColumn bool CODEC(ZSTD(1)),
+) ENGINE ReplacingMergeTree
+ORDER BY (tagKey, tagType, dataType, isColumn);
```
srikanthccv, line 7: Does this need to keep the values forever? Can it have the same maximum TTL as the index table?

exporter/clickhousetracesexporter/migrations/000023_span_attributes_mv.up.sql
makeavish: It's possible that there are two keys with the same name but one is column and the other is not, as we discussed we don't want to put restrictions on insertion of such keys and have to handle them on query builder (while querying)

exporter/clickhousetracesexporter/writer.go
makeavish: We can merge the `writeTagKeyBatch` and `writeTagBatch` to reduce iterations.

exporter/clickhousetracesexporter/migrations/000024_span_attributes_keys.up.sql
makeavish: But it's expected that keys won't be more than 10s of thousands, keys table of logs also doesn't have a TTL

exporter/clickhousetracesexporter/migrations/000023_span_attributes_mv.up.sql
srikanthccv: I think this will cause issues with materialising the column in future.

srikanthccv (approved):

PR #127, opened 2023-04-19
chore: merge tagKeyBatch and tagBatch writer funcs

makeavish:


exporter/clickhousetracesexporter/writer.go
```diff
@@ -163,17 +163,10 @@ func (w *SpanWriter) writeBatch(batch []*Span) error {
...
 			logBatch := batch[:int(math.Min(10, float64(len(batch))))]
-			w.logger.Error("Could not write a batch of spans to tag table: ", zap.Any("batch", logBatch), zap.Error(err))
-			return err
-		}
-	}
-	if w.attributeKeyTable != "" {
-		if err := w.writeTagKeyBatch(batch); err != nil {
-			logBatch := batch[:int(math.Min(10, float64(len(batch))))]
-			w.logger.Error("Could not write a batch of spans to tag key table: ", zap.Any("batch", logBatch), zap.Error(err))
+			w.logger.Error("Could not write a batch of spans to tag/tagKey tables: ", zap.Any("batch", logBatch), zap.Error(err))
```
srikanthccv, line 169: Put `zap.Error(err)` first so the user can see the error first before they have to scroll past the batch dump. How does logging a batch of spans help with the error here?

exporter/clickhousetracesexporter/writer.go
```diff
@@ -251,17 +244,34 @@ func (w *SpanWriter) writeIndexBatch(batchSpans []*Span) error {
...
 		logBatch := batchSpans[:int(math.Min(10, float64(len(batchSpans))))]
 		w.logger.Error("Could not prepare batch for span attributes table: ", zap.Any("batch", logBatch), zap.Error(err))
 		return err
 	}
-
+	tagKeyStatement, err := w.db.PrepareBatch(ctx, fmt.Sprintf("INSERT INTO %s.%s", w.traceDatabase, w.attributeKeyTable))
+	if err != nil {
+		logBatch := batchSpans[:int(math.Min(10, float64(len(batchSpans))))]
+		w.logger.Error("Could not prepare batch for span attributes key table: ", zap.Any("batch", logBatch), zap.Error(err))
+		return err
```
srikanthccv, line 257: You are already logging in the caller - avoid logging again and just return the err up in the call chain.

exporter/clickhousetracesexporter/writer.go
```diff
@@ -251,17 +244,34 @@ func (w *SpanWriter) writeIndexBatch(batchSpans []*Span) error {
...
 		for _, spanAttribute := range span.SpanAttributes {
+
+			err = tagKeyStatement.Append(
+				spanAttribute.Key,
+				spanAttribute.TagType,
+				spanAttribute.DataType,
+				spanAttribute.IsColumn,
+			)
+			if err != nil {
+				w.logger.Error("Could not append span to tagKey Statement to batch: ", zap.Object("span", span), zap.Error(err))
```
srikanthccv, line 269: Same here

srikanthccv (commented):
LGTM. Some question about the usefulness of logging big objects in the key/values context and duplicate logging.

exporter/clickhousetracesexporter/writer.go
makeavish: > Put zap.Error(err) first so the user can see the error first before they have to scroll past the batch dump.

Done.

> How does logging a batch of spans help with the error here?

Debugging is easier if we know the spans. 
As we write to CH in batch and can't identify individual spans which caused error, so we need to log the whole batch if there's an error.

exporter/clickhousetracesexporter/writer.go
makeavish: Yes

srikanthccv (commented):
Can you move the error `zap.Error ` first so that error can be seen immediately before gaint marshalled spans data?

It is still not clear as how logging the batch of 10 marshalled spans helps with `PrepareBatch` error. It doesn't even include the column names which could potentially cause the issue. If we could dig deep into the error which caused the issue earlier and get the root cause of it that would be great. That could be addressed separately.

srikanthccv (approved):
I still believe this unnecessarily logs for `.PrepareBatch()` and for `.send()`, where the reason is still speculated. If it were to be serialisation, when the error occurs, it continuously occurs unless the serialisation issue is fixed. The marshalling and heavy logging deteriorates the collector over the process lifetime.

PR #134, opened 2023-05-23
Update clickhouse_exporter.go to handle both String http.status_code

ryansmartpadpro:
This changes http.status_code to be handled the same way that grpc.status_code in that both string and int values are valid


srikanthccv (approved):

makeavish (approved):

ryansmartpadpro:
@makeavish is there any changes needed before this can be merged?

PR #136, opened 2023-06-02
 chore: optimize multi architecture build

prashant-shahi:
- make targets for creation of static binaries
- optimize multi architecture image build flow by building artifacts outside of container
- update dependent workflows
- include intermediate `.build` folder in .gitignore file

Old PR: #93

Signed-off-by: Prashant Shahi <prashant@signoz.io>


Makefile
```diff
@@ -31,7 +33,18 @@ test:
 
 .PHONY: build
 build:
-	GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o signoz-collector ./cmd/signozcollector
+	CGO_ENABLED=1 go build -tags timetzdata -o .build/${GOOS}-${GOARCH}/signoz-collector -ldflags "-linkmode external -extldflags '-static' -s -w ${LD_FLAGS}" ./cmd/signozcollector
```
srikanthccv, line 36: Was there a reason why `CGO_ENABLED` changed?

Makefile
```diff
@@ -31,7 +33,18 @@ test:
...
+
+.PHONY: amd64
+amd64:
+	make GOARCH=amd64 build
+
+.PHONY: arm64
+arm64:
+	make CC=aarch64-linux-gnu-gcc GOARCH=arm64 build
```
srikanthccv, line 44: This requires the gcc installed on system and problem for working on any machine that doesn't use gnu gcc.

prashant-shahi: yeah, you are right.

However, this is only meant to be used by us for cross-compiling for building multi-arch docker image.

For users wanting to build binary for their own machine, they can still use `make build`.

Makefile
prashant-shahi: Till Go `1.19`, `CGO_ENABLED` is enabled by default.
Since we were using alpine base image for both build time and run time,
we did not face any issues even though we had `ENV CGO_ENABLED=0` hardcoded in Dockerfile.

In case we set to `CGO_ENABLED` is false and build the static binary.
We get the following `core dumped` error when trying to run it.

```
[1]    2090581 trace trap (core dumped)  .build/linux-amd64/signoz-collector
```

The default behaviour of cgo is altered from Go 1.20. Hence, it is safer to enable
it on our own rather than relying on the defaults. Ref: https://tip.golang.org/doc/go1.20#cgo

srikanthccv (approved):

srikanthccv:
Now that you published binaries, can you try building the docker image using make on systems that don't have gnu-gcc?

prashant-shahi:
> Now that you published binaries, can you try building the docker image using make on systems that don't have gnu-gcc?

If you meant to build multi-arch docker image, it would still require `gnu-gcc`. Since we are now building binaries outside of the docker container.

However, I don't think normally you would have to manually do that. You can either use `main` tag for merged PR, or create release candidate tag.

In case of amd/arm docker image, you can still do that without any issue in respective amd/arm machine.

srikanthccv:
>In case of amd/arm docker image, you can still do that without any issue in respective amd/arm machine.

No, this isn't working. I am on a branch X and want to build a docker image with the current changes. The make targets no longer work on the Apple machine for `arm64`.

prashant-shahi:
>  The make targets no longer work on the Apple machine for arm64.

You can build it on the same machine and use it

In case you need to share it with others and multi architecture image is a "must have", you can use staging env to build it.

prashant-shahi:
I understand the trade-off with building outside the container.

However, we should also note about the build time improved of `push-workflow` by 4.5-6 times.
Before it used to take 45-60 minutes, now it takes around 10 minutes.

prashant-shahi:
@srikanthccv addressed in #139

PR #140, opened 2023-06-13
chore: updated otel col to v0.79.0

makeavish:


srikanthccv:
Test the nginx receiver breaking change once if it affects our users. LGTM otherwise.

makeavish:
Checked the nginx receiver change, it's under a feature flag that our users need not use until it replaces original metric.

srikanthccv (approved):

Issue #147, opened 2023-06-23
`hasError` must not be set true for server kind spans with 4xx


srikanthccv:
https://github.com/SigNoz/signoz-otel-collector/blob/8decaf03afdfcd7d7167738642972f2f2afc7f45/exporter/clickhousetracesexporter/clickhouse_exporter.go#L153-L164

PR #153, opened 2023-06-26
fix: hasError not set true

makeavish:
for server kind spans with 4xx errors

fixes(https://github.com/SigNoz/signoz-otel-collector/issues/147)


exporter/clickhousetracesexporter/clickhouse_exporter.go
```diff
@@ -157,7 +157,7 @@ func populateOtherDimensions(attributes pcommon.Map, span *Span) {
 			if err == nil && statusString != 0 {
 				statusInt = int64(statusString)
 			}
-			if statusInt >= 400 {
+			if statusInt >= 500 || (statusInt >= 400 && span.Kind != 2) {
```
srikanthccv, line 160: There is a gRPC as well. Please update that.

https://github.com/SigNoz/signoz-otel-collector/blob/8decaf03afdfcd7d7167738642972f2f2afc7f45/exporter/clickhousetracesexporter/clickhouse_exporter.go#L198-L207

makeavish: For gRPC apparently it's not very clear which are server side errors.
Go through these docs, you will see that most status codes are shared among client and server: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

makeavish:
> As discussed on 26th Jun, Please rely on the span status code for hasError.

@srikanthccv Updated code to only rely on span status code for has error. [Here](https://github.com/SigNoz/signoz-otel-collector/blob/926fba7740d54250942ad39dd289b56cd27b8985/exporter/clickhousetracesexporter/clickhouse_exporter.go#L367) we are checking if span status code is 2 to set hasError to true

srikanthccv:
Can you use the symbol `ptrace.StatusCodeError` from pdata instead of 2 in the linked code?

makeavish:
> Can you use the symbol `ptrace.StatusCodeError` from pdata instead of 2 in the linked code?

Updated

srikanthccv (approved):

Issue #157, opened 2023-07-10
[Traces] Usge collection should happen at newStructuredSpan


nityanandagohain:
As of now due to tagMap and stringMap the size is double, calculate it at newStructuredSpan function instead.

PR #158, opened 2023-07-16
fix: update usage collection to remove deprecated param

makeavish:


nityanandagohain (changes requested):
This is incorrect, its doing the same thing i.e size of entire tracemodel is added again.

Please use the solution mentioned here https://github.com/SigNoz/signoz-otel-collector/issues/157#issue-1797975584

nityanandagohain (changes requested):
Let's calculate at the receiving level itself https://github.com/SigNoz/signoz-otel-collector/issues/157#issue-1797975584 ?

nityanandagohain (approved):

srikanthccv:
Marshalling and JSON, especially, are very expensive in CPU resource usage. The seemingly simple change made it almost 2x what it used to be, and it will degrade the collector capability when the load is non-trivial. I would have looked for other approaches. 

Side note: @nityanandagohain, it's good to write down why you went from requested changes to approved again without any comments. I am assuming you had some discussion which is fine but others wouldn't know anything about it.

nityanandagohain:
@makeavish lets open a new issue so that we keep the the resource usage in account and make changes once we find a better solution.

PR #183, opened 2023-10-06
feat: boolean attribute and dot in attribute names

nityanandagohain:
* Adds support for boolean attribute.
* Doesn't convert `.` to `_` in attribute names.

Fixes https://github.com/SigNoz/signoz/issues/3553


nityanandagohain:
Not relevant as of now since we are not going with `.`

PR #184, opened 2023-10-06
feat: templatize migrations

### LINKED ISSUE #180: Question: Is there any way to choose cluster name?
From what I see in code cluster name is const == 'cluster'. Maybe there is some workaround?

dhawal1248:
Fixes #180


nityanandagohain (changes requested):
Make commands are not working as the dockerfiles are not updated.

dhawal1248:
> There are still some instances of hard-coded cluster name `ON CLUSTER cluster`. Please make sure all of them are templatized.

Yes @srikanthccv , i found a few missed occurences of on CLUSTER cluster while testing, have changed all of them.There also a few occurences of Distributed("cluster", ...) i had missed..have changed them too

cmd/signozcollectormigrator/migrate.go
```diff
@@ -0,0 +1,99 @@
...
+	f := pflag.NewFlagSet("Collector Migrator CLI Options", pflag.ExitOnError)
+
+	f.Usage = func() {
+		fmt.Println(f.FlagUsages())
+		os.Exit(0)
+	}
+
+	f.String("dsn", "", "Clickhouse DSN")
+	f.String("cluster-name", "cluster", "Cluster name to use while running migrations")
+	f.Bool("multi-node-cluster", false, "True if the dsn points to a multi node clickhouse cluster, false otherwise. Defaults to false.")
```
srikanthccv, line 40: Why do we still have this `multi-node-cluster`?

cmd/signozcollectormigrator/migrate.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if dsn == "" {
+		logger.Fatal("dsn is a required field")
+	}
+
+	// set cluster env so that golang-migrate can use it
+	// the value of this env would replace all occurences of {{.SIGNOZ_CLUSTER}} in the migration files
+	os.Setenv("SIGNOZ_CLUSTER", clusterName)
+
+	managerConfig := migrationmanager.NewConfig(
```
srikanthccv, line 82: (nit): common practice is to expose the config options and let the module consumer create the config object directly instead of the `NewConfig` func.

migrationManager/migrators/baseMigrator/migrator.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+
+	// drop schema migrations table if running in docker multi node cluster mode so that migrations are run on new nodes
+	if m.Cfg.IsMultiNodeCluster {
+		err := m.dropSchemaMigrationsTable(ctx, database)
+		if err != nil {
+			return err
+		}
+	}
```
srikanthccv, line 50: Same question, why does this need to drop?

cmd/signozcollectormigrator/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# create a non-root user for running the migrator
+USER ${USER_UID}
+
+# copy the binaries from the multi-stage build
+COPY .build/${TARGETOS}-${TARGETARCH}/signoz-collector-migrator /signoz-collector-migrator
+
+# run the binary as the entrypoint and pass the default dsn as a flag
+ENTRYPOINT [ "/signoz-collector-migrator" ]
+CMD ["--dsn", "tcp://host.docker.internal:9000"]
```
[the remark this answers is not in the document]
dhawal1248: @srikanthccv this change was intentional..since we are expecting people to run images built by this dockerfile on containers, localhost:9000 might not point to a clickhouse server..in such a case host.docker.internal:9000 is a better default value as it uses the docker host's port 9000 which is much more likely to be bound to a clickhouse server(either running natively or port forwarded from a container)

cmd/signozcollectormigrator/migrate.go
```diff
@@ -0,0 +1,99 @@
...
+	config := zap.NewProductionConfig()
+	config.EncoderConfig.EncodeLevel = zapcore.LowercaseLevelEncoder
+	config.EncoderConfig.TimeKey = "timestamp"
+	config.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	logger, err := config.Build()
+	if err != nil {
+		log.Fatalf("Failed to initialize zap logger %v", err)
+	}
+	// replace global logger
+	zap.ReplaceGlobals(logger)
```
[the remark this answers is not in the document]
dhawal1248: @srikanthccv yeah i agree, we should move away from this
Can we add a todo here and move on with releasing this as is? I will make sure i pick it up later and fix it

cmd/signozcollectormigrator/migrate.go
dhawal1248: Yeah i agree @srikanthccv i had used the same thing earlier but then i had to create a manager config as the parameters were increasing
I wanted to create a singular struct with the config and wanted to use it both in the manager and the individual migrators for metrics, etc..there were couple of things i was trying to do via the newConfig func - 
- i wanted the manager config to be a superset of the migrator config(for being able to easily add usecases like running migrations only for one of logs/metrics/traces)
- i wanted to keep cmd's migrate.go only dependent on the migrationmanager's manager.go file.Didn't want to import migrator config inside the migrate.go file
But looking back the number of arguments to New reaching 5 was kind of my threshold for adding a separate config struct for manager..now that we have removed the multi node flag, it brings the number of arguements to the New func to 4 which makes this whole config struct look like an overkill to me
For now, i have removed the manager's Config struct and am passing 4 arguments to manager's new func to create a new manager

cmd/signozcollectormigrator/migrate.go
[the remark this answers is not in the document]
dhawal1248: I am a little hesitant to change this as we have already tested it quite a bit so dont want to make a lot of changes..passing the logger through args would need changes in a lot of places

cmd/signozcollectormigrator/migrate.go
srikanthccv: You misunderstood my point, it's fine to have Config objects and have the modules initialising them. But the `NewConfig` function to create a `Config` object is uncommon.

cmd/signozcollectormigrator/migrate.go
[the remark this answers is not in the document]
srikanthccv: Please create a separate issue, although I don't believe we will pick it up and fix it because I heard that lie we tell ourselves too often.

cmd/signozcollectormigrator/Dockerfile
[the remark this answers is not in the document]
srikanthccv: This work is more important in other environments than the default case it tries to handle. You may have had this convenient for testing, but in the real world, this will be used in k8s environments 90% of the time. For this reason, I believe this docker networking specific _running in a container trying to access host ClickHouse_ shouldn't be the default.

Makefile
```diff
@@ -73,6 +75,24 @@ build-signoz-collector:
 		--no-cache -f cmd/signozcollector/Dockerfile --progress plain \
 		--tag $(REPONAME)/$(IMAGE_NAME):$(DOCKER_TAG) .
 
+.PHONY: build-signoz-schema-migrator
+build-signoz-schema-migrator:
+	@echo "------------------"
+	@echo  "--> Build signoz collector migrator docker image"
```
srikanthccv, line 81:
```suggestion
	@echo  "--> Build signoz schema migrator docker image"
```

Makefile
```diff
@@ -73,6 +75,24 @@ build-signoz-collector:
...
+	@echo "------------------"
+	docker build --build-arg TARGETPLATFORM="linux/amd64" \
+		--no-cache -f cmd/signozschemamigrator/Dockerfile --progress plain \
+		--tag $(REPONAME)/$(MIGRATOR_IMAGE_NAME):$(DOCKER_TAG) .
+
+.PHONY: build-and-push-signoz-schema-migrator
+build-and-push-signoz-schema-migrator:
+	@echo "------------------"
+	@echo  "--> Build and push signoz collector migrator docker image"
```
srikanthccv, line 90:
```suggestion
	@echo  "--> Build and push signoz schema migrator docker image"
```

cmd/signozschemamigrator/migrate.go
```diff
@@ -0,0 +1,87 @@
...
+		log.Fatalf("Failed to initialize zap logger %v", err)
+	}
+	// replace global logger
+	// TODO(dhawal1248): move away from global logger
+	zap.ReplaceGlobals(logger)
+}
+
+func main() {
+	logger := zap.L().With(zap.String("component", "migrate cli"))
+	f := pflag.NewFlagSet("Collector Migrator CLI Options", pflag.ExitOnError)
```
srikanthccv, line 32:
```suggestion
	f := pflag.NewFlagSet("Schema Migrator CLI Options", pflag.ExitOnError)
```

example/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: "3.8"
+
+services:
+  otel-collector-migrator:
```
srikanthccv, line 4:
```suggestion
  schema-migrator:
```

example/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - /var/lib/docker/containers:/var/lib/docker/containers:ro
+      - /var/log:/var/log:ro
+    # required to read logs from /var/lib/docker/containers/
+    ports:
+      - "1888" # pprof extension
+      - "8888" # Prometheus metrics exposed by the collector
+      - "13133" # health_check extension
+      - "55679" # zpages extension
+    depends_on:
+      otel-collector-migrator:
```
srikanthccv, line 32:
```suggestion
      schema-migrator:
```

example/example-config.yaml
```diff
@@ -0,0 +1,142 @@
...
+extensions:
+  health_check:
+  pprof:
+  zpages:
+  memory_ballast:
+    size_mib: 1000
+exporters:
+  clickhousetraces:
+    datasource: tcp://signoz-host:9000/?database=signoz_traces
+    migrations: exporter/clickhousetracesexporter/migrations
```
srikanthccv, line 112: Remove this config option here https://github.com/SigNoz/signoz-otel-collector/blob/main/exporter/clickhousetracesexporter/config.go#L25

srikanthccv (approved):
Non-blocking comments

example/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: "3.8"
+
+services:
+  otel-collector-migrator:
+    image: signoz/signoz-schema-migrator:templatize-migrations
```
prashant-shahi, line 5: update this to a tagged version after one is created.

cmd/signozcollectormigrator/migrate.go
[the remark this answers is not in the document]
dhawal1248: https://github.com/SigNoz/signoz-otel-collector/issues/197

migrationmanager/migrators/logs/migrations/000005_attribute_rename.down.sql
```diff
@@ -1,32 +1,32 @@
-RENAME TABLE IF EXISTS signoz_logs.logs_attribute_keys TO signoz_logs.logs_atrribute_keys on CLUSTER cluster;
+RENAME TABLE IF EXISTS signoz_logs.logs_attribute_keys TO signoz_logs.logs_atrribute_keys ON CLUSTER {{.SIGNOZ_CLUSTER}};
```
nityanandagohain, line 1: Please add a line from here https://github.com/SigNoz/signoz-otel-collector/blob/5758bc46545f82997beaf43faa3c4cfa7f428584/exporter/clickhouselogsexporter/migrations/000005_attribute_rename.up.sql#L2 as it was added recently.

nityanandagohain (changes requested):
One change, rest looks good to me

prashant-shahi (approved):

PR #187, opened 2023-10-09
fix: migration reapply issue fixed for logs

### LINKED ISSUE #186: Logs: fix migration 5
If 5th migration is applied and we reapply the migration it fails

nityanandagohain:
Fixes #186

### DIFF
--- a/exporter/clickhouselogsexporter/migrations/000005_attribute_rename.up.sql
+++ b/exporter/clickhouselogsexporter/migrations/000005_attribute_rename.up.sql
@@ -1,3 +1,6 @@
+-- We are dropping logs_attribute_keys as we are renaming logs_atrribute_keys in the next line
+DROP TABLE IF EXISTS signoz_logs.logs_attribute_keys on CLUSTER cluster;
+
 RENAME TABLE IF EXISTS signoz_logs.logs_atrribute_keys TO signoz_logs.logs_attribute_keys on CLUSTER cluster;
 
 


srikanthccv:
Can you edit the existing migration files without getting into checksum issues on instances where they are already applied? Is this needed, given we will have the migrations issue fixed with Dhawal's work soon?

nityanandagohain:
So it fixes the cases where if the migration is already applied and we reapply the migration then it won't cause any issue. This is independent of dhawals work.

srikanthccv:
My first question is not answered. It's not a good idea to edit the existing migrations. It only makes things worse. https://edgeguides.rubyonrails.org/active_record_migrations.html#changing-existing-migrations, this is a RoR guide (I could quickly find a link) but applies across languages/frameworks. If it is a brand new first release, then there is no issue, but there are production users and OSS users who have run it already. 

Coming to the second question, does reapplying alone cause this issue? Or does the reapply after the earlier failed migration cause the issue? If it is the latter, then we can assume Dhawal's work will eliminate the possibility of migration failures, and we don't need to do this. If it is the former, the first question is relevant.

nityanandagohain:
it's the former i.e reapplying the migration causes the issue. Ex- If things are fine and I delete my schema migrations table and restart my collecotor then it won't start and this PR fixes that.

For this case I tested and it doesn't throw any issue as if the migrations are already appliead then this won't even run. Only when all the migrations are rerun this needs to be present.

srikanthccv (approved):
Thanks for answering my questions.

PR #190, opened 2023-10-10
tracesexporter: Allow multiple `datasources` in configuration, and round-robin load-balance amongst them

ELLIOTTCABLE:
This is my first-ever time touching anything Go related, so please, definitely go over my patch with a fine-tooth comb … very open to any sort of criticism!

This replaces the singular `db *clickhouse.Conn` values throughout the `clickhousetracesexporter` with an array of `conns`. Any db usage is routed through a simple round-robin load-balancer, with the exception if the initialization, which should stay on a single initial connection from the pool.

I've got a few questions, will attach them as a self-review below.

### DIFF
--- a/exporter/clickhousetracesexporter/clickhouse_factory.go
+++ b/exporter/clickhousetracesexporter/clickhouse_factory.go
@@ -33,12 +33,12 @@ import (
 
 // Factory implements storage.Factory for Clickhouse backend.
 type Factory struct {
-	logger     *zap.Logger
-	Options    *Options
-	db         clickhouse.Conn
-	archive    clickhouse.Conn
-	datasource string
-	makeWriter writerMaker
+	logger        *zap.Logger
+	Options       *Options
+	conns         []clickhouse.Conn
+	archive_conns []clickhouse.Conn
+	datasource    string
+	makeWriter    writerMaker
 }
 
 // Writer writes spans to storage.
@@ -55,7 +55,7 @@ var (
 )
 
 // NewFactory creates a new Factory.
-func ClickHouseNewFactory(migrations string, datasource string, dockerMultiNodeCluster bool) *Factory {
+func ClickHouseNewFactory(migrations string, datasources []string, dockerMultiNodeCluster bool) *Factory {
 	writeLatencyDistribution := view.Distribution(100, 250, 500, 750, 1000, 2000, 4000, 8000, 16000, 32000, 64000, 128000, 256000, 512000)
 
 	writeLatencyView := &view.View{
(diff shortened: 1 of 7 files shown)


exporter/clickhousetracesexporter/clickhouse_factory.go
```diff
@@ -78,54 +77,64 @@ func ClickHouseNewFactory(migrations string, datasource string, dockerMultiNodeC
...
+func (f *Factory) selectConn() clickhouse.Conn {
+	if len(f.conns) > 1 {
+		f.conns = append(f.conns[1:], f.conns[0])
+	}
+	return f.conns[0]
+}
+
 // Initialize implements storage.Factory
-func (f *Factory) Initialize(logger *zap.Logger) error {
+func (f *Factory) Initialize(logger *zap.Logger) (clickhouse.Conn, error) {
```
ELLIOTTCABLE, line 89: Is there a clear way in Go to indicate _what_ this return-value is, semantically? (I name the return-value `init_conn` below, but idk if folks are going to reflexively check the source …)

I'd rather indicate in the type/signature that the user shouldn't use the returned `conn` for anything other than initialization …

exporter/clickhousetracesexporter/options.go
```diff
@@ -48,7 +48,7 @@ const (
 
 const (
 	suffixEnabled         = ".enabled"
-	suffixDatasource      = ".datasource"
+	suffixDatasources     = ".datasources"
```
ELLIOTTCABLE, line 51: I'm unclear on the best path forward here, re: naming, if my changes are adopted:

1. Continue to use `datasource:`, even though it can now be plural;
2. Switch names to be more semantic, but break backwards compat;
3. Match the other exporter(s) and use `DSN:` or even `DSNs:`?

exporter/clickhousetracesexporter/options.go
```diff
@@ -188,6 +203,19 @@ func (opt *Options) AddFlags(flagSet *flag.FlagSet) {
...
+
+func (ssv *stringSliceValue) String() string {
+	return fmt.Sprintf("%v", *ssv.slice)
+}
+
+func (ssv *stringSliceValue) Set(value string) error {
+	*ssv.slice = append(*ssv.slice, value)
+	return nil
+}
+
```
ELLIOTTCABLE, line 218: This is mostly via Google 🙈 — I'm not really familiar with `flag`; is this the appropraite way to get the desired `--datasource foo --datasource bar` behaviour?

exporter/clickhousetracesexporter/clickhouse_factory.go
```diff
@@ -78,54 +77,64 @@ func ClickHouseNewFactory(migrations string, datasource string, dockerMultiNodeC
...
 }
 
+// Round-robin connection-selection strategy.
+func (f *Factory) selectConn() clickhouse.Conn {
+	if len(f.conns) > 1 {
+		f.conns = append(f.conns[1:], f.conns[0])
+	}
+	return f.conns[0]
+}
+
```
ELLIOTTCABLE, line 87: This is, currently, not thread-safe. I'm not sure whether parallelism is utilized in `signoz-otel-collector`; should I include a mutex against the list of collectors or something like that?

exporter/clickhousetracesexporter/writer.go
```diff
@@ -86,10 +86,19 @@ func NewSpanWriter(options WriterOptions) *SpanWriter {
...
 }
 
-func (w *SpanWriter) writeIndexBatch(batchSpans []*Span) error {
+// Round-robin connection-selection strategy.
+func (w *SpanWriter) selectConn() clickhouse.Conn {
+	if len(w.conns) > 1 {
+		w.conns = append(w.conns[1:], w.conns[0])
+	}
+	return w.conns[0]
+}
```
ELLIOTTCABLE, line 95: (and is there a way I can centralize/dedupe `Writer.selectConn` and `Factory.selectConn`, since they're identical?)

Issue #107, opened 2023-03-01
span metrics insertion from the same collector instance


srikanthccv:
the collector-metrics scrape is prone to issues, and users get worried when they don't see charts for service overview which is built based on the span metrics data. we could potentially run a go routine in span metrics processor that periodically inserts the data into clickhouse.

PR #201, opened 2023-10-20
Update signozspanmetricsprocessor to remove dependency on prometheus …

srikanthccv:
…exporter

Fixes https://github.com/SigNoz/signoz-otel-collector/issues/107

Contributes to an effort to use delta temporality for span metrics. There are some cases where Cumulative shines, such as 

1. A failed export for delta is failed, but a failed export for Cumulative could be fixed by interpolation/extrapolation (not fully accurate as a real value, but a good approximation). 
2. Double entry of samples from retries wouldn't be an issue with Cumulative because we take the max of values from the time range, whereas the sum in delta might double count

However, unlike purpose-built systems like Prometheus, We don't exploit any of the characteristics or handle challenges that come with it. There is no reset handling, no interpolation/extrapolation, and more, mainly because ClickHouse doesn't have a way to handle them. We incline towards Delta because it's easier to handle; there is no need to worry about stale metrics because they are flushed for each interval; it is faster, less prone to memory issues, and more accurate because we don't need to deal with the complication that arises from resets etc...


srikanthccv:
Please review

nityanandagohain (approved):

makeavish (approved):
We should add more comments/descriptions if we are also committing unrelated changes.

processor/signozspanmetricsprocessor/config.go
```diff
@@ -80,6 +80,9 @@ type Config struct {
 
 	// skipSanitizeLabel if enabled, labels that start with _ are not sanitized
 	skipSanitizeLabel bool
+
+	// MetricsEmitInterval is the time period between when metrics are flushed or emitted to the configured MetricsExporter.
+	MetricsFlushInterval time.Duration `mapstructure:"metrics_flush_interval"`
```
ankitnayan, line 85: This interval is used to flush the delta metrics being calculated in-memory? I remember the initial implementation I had caused increase in writes to clickhouse as the calculated metrics were designed to be written immediately causing heavy DB write/s. I hope we should not face that with this configuration.

Another thing, this flush interval was there before the introduction of  delta temporality too? I want to understand why was it there before if we used to scrape things like prometheus and not push at flushInterval like defined here

srikanthccv: >This interval is used to flush the delta metrics being calculated in-memory?

Yes

> I remember the initial implementation I had caused increase in writes to clickhouse as the calculated metrics were designed to be written immediately causing heavy DB write/s. I hope we should not face that with this configuration.

Yes, this accumulates the values for each interval and then sends the data to the exporter.

> this flush interval was there before the introduction of delta temporality too?

No, it wasn't there. The previous design had the cumulative value which never resets but the Prometheus was configured to read the cumulative value every 1 minutes (or scrape interval).

ankitnayan: got it thanks.

PR #212, opened 2023-10-26
fix: ingest bool attrs

### LINKED ISSUE #209: Bool attributes are dropped

nityanandagohain:
Fixes #209 

* Previously we directly stored bool values as as string.
* When we added tag attributes support that time this bug got introduct when the attribute wasn't stored in with log line.
* Added changes so that the bool values are stored.

Suggestes and querying work as expected.

### DIFF
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -342,8 +342,7 @@ func attributesToSlice(attributes pcommon.Map, forceStringValues bool) (response
 				response.FloatValues = append(response.FloatValues, v.Double())
 			case pcommon.ValueTypeBool:
 				// add boolValues in future if it is required
-				response.BoolKeys = append(response.BoolKeys, formatKey(k))
-				// store it as a string in clickhouse
+				// store it as a string in clickhouse for now
 				response.StringKeys = append(response.StringKeys, formatKey(k))
 				response.StringValues = append(response.StringValues, v.AsString())
 			default: // store it as string
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -343,6 +343,9 @@ func attributesToSlice(attributes pcommon.Map, forceStringValues bool) (response
 			case pcommon.ValueTypeBool:
 				// add boolValues in future if it is required
 				response.BoolKeys = append(response.BoolKeys, formatKey(k))
+				// store it as a string in clickhouse
+				response.StringKeys = append(response.StringKeys, formatKey(k))
+				response.StringValues = append(response.StringValues, v.AsString())
 			default: // store it as string
 				response.StringKeys = append(response.StringKeys, formatKey(k))
 				response.StringValues = append(response.StringValues, v.AsString())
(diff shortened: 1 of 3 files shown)


exporter/clickhouselogsexporter/exporter.go
```diff
@@ -342,7 +342,9 @@ func attributesToSlice(attributes pcommon.Map, forceStringValues bool) (response
 				response.FloatValues = append(response.FloatValues, v.Double())
 			case pcommon.ValueTypeBool:
 				// add boolValues in future if it is required
-				response.BoolKeys = append(response.BoolKeys, formatKey(k))
+				// store it as a string in clickhouse for now
+				response.StringKeys = append(response.StringKeys, formatKey(k))
+				response.StringValues = append(response.StringValues, v.AsString())
```
srikanthccv, line 347: > Added changes so that the bool values are stored


I am trying to understand why/where are they dropped currently and why do we need to store them as string?

srikanthccv: Also this had related bool support and different implementation https://github.com/SigNoz/signoz-otel-collector/pull/183

nityanandagohain: So they were dropped because previously we stored them in the string key values array. But after adding tag attributes we added bool but it was not actually stored with the log line.

yeah adding the materialized table will be better with new column. Let me make the changes.

migrationmanager/migrators/logs/migrations/000008_add_bool.down.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE signoz_logs.logs ON CLUSTER cluster DROP column IF EXISTS attributes_bool_key;
```
srikanthccv, line 1: The cluster shouldn't be hardcoded. It should be template env.

prashant-shahi (approved):

migrationmanager/migrators/logs/migrations/000008_add_bool.up.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} add column IF NOT EXISTS attributes_bool_key Array(String) CODEC(ZSTD(1));
+
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} add column IF NOT EXISTS attributes_bool_value Array(Bool) CODEC(ZSTD(1));
+
+ALTER TABLE signoz_logs.distributed_logs ON CLUSTER {{.SIGNOZ_CLUSTER}} add column IF NOT EXISTS attributes_bool_key Array(String) CODEC(ZSTD(1));
+
+ALTER TABLE signoz_logs.distributed_logs ON CLUSTER {{.SIGNOZ_CLUSTER}} add column IF NOT EXISTS attributes_bool_value Array(Bool) CODEC(ZSTD(1));
+
+CREATE MATERIALIZED VIEW IF NOT EXISTS  attribute_keys_bool_final_mv ON CLUSTER {{.SIGNOZ_CLUSTER}} TO signoz_logs.logs_attribute_keys AS
```
srikanthccv, line 9:
```suggestion
CREATE MATERIALIZED VIEW IF NOT EXISTS  signoz_logs.attribute_keys_bool_final_mv ON CLUSTER {{.SIGNOZ_CLUSTER}} TO signoz_logs.logs_attribute_keys AS
```

srikanthccv (approved):

PR #215, opened 2023-10-30
chore: fixes for dirty migration issue

dhawal1248:
- add some logs around env's to be able to debug better
- bump golang-migrate version


cmd/signozschemamigrator/migrate.go
```diff
@@ -72,7 +73,22 @@ func main() {
...
+	clusterNameFromEnv := ""
+	for _, kvp := range os.Environ() {
+		kvParts := strings.SplitN(kvp, "=", 2)
+		if kvParts[0] == "SIGNOZ_CLUSTER" {
+			clusterNameFromEnv = kvParts[1]
+			break
+		}
+	}
+	if clusterName == "" {
+		logger.Fatal("Failed to set env var SIGNOZ_CLUSTER")
```
srikanthccv, line 89: `os.Setenv` returns an error when it fails to set or something else happens. It is also the most reliable info to check if the call to `Setenv` was successful.

srikanthccv: Since we are using the fork of the migrate and can make any changes, adding some verbose logs close to the template exec in the SigNoz/golang-migrate is better.

srikanthccv: Is there a reason why this runs in go routine https://github.com/SigNoz/golang-migrate/blob/7e92866dfed375fc6f0a9190d7b2324f8f6cbdc3/template.go#L41-L44 and return errors are ignored?

srikanthccv (commented):
I think there should be more verbose logs in golang-migrate codebase than here to fully understand the problem.

cmd/signozschemamigrator/migrate.go
dhawal1248: Not sure @srikanthccv, this is the library's implementation

srikanthccv: How's that a library implementation? We added the template part.

dhawal1248: I'm sorry i didn't phrase it properly..it is a library fork's implementation which we are using

dhawal1248: > Since we are using the fork of the migrate and can make any changes, adding some verbose logs close to the template exec in the SigNoz/golang-migrate is better.

I have added some logs to near exec in the migrate fork too @srikanthccv 
i'm also keeping the logs here in migrate.go for now..it would be duplicate logs but we can remove them once this issue is figured out

PR #221, opened 2023-11-03
fix: eclude chi and otelcol metrics from usage reporting

nityanandagohain:
Excludes `chi_` and `otelcol_` metrics from usage collection


makeavish (approved):

prashant-shahi (approved):

srikanthccv:
I have some questions about this. Please hold.

srikanthccv:
@prashant-shahi why do we allow our collectors/clickhouse internal monitoring metrics to be written to user instances? https://signoz-team.slack.com/archives/C050MERV57W/p1699156207666079?thread_ts=1698929179.596819&cid=C050MERV57W

srikanthccv (approved):
I will approve this for now but I neither like the idea the writing our internal monitoring data to the users table in the first place (why would they want to look at our CHI metrics?) nor the blanket exclusion based on name because that doesn't guarantee they are only generated by us unlike the span metrics.

prashant-shahi:
> why do we allow our collectors/clickhouse internal monitoring metrics to be written to user instances?

@srikanthccv This is only specific to the self-hosted SigNoz and not Cloud, where the internal collector/chi metrics are collected and stored in their clickhouse.

srikanthccv:
I think this should go. We had incorrect cloud docs which enabled the metrics of otelcol-binary. The "proper" fix can be made later. Imagine someone running binary in 200+ machines and sending these metrics that they have no idea about or any use.

Issue #222, opened 2023-11-03
Clickhouse connections are not released


makeavish:
When clickhouse max connections are reached all the consumers start facing `acquire conn timeout` errors and no data is written to Clickhouse.
That shouldn't be the case and data should continue to be written. Maybe there's a deadlock due to which writes are not happening.

nityanandagohain:
This is the example code I used to reproduce

Here same acquire connection issue is produced but as it retries and older connection are released the retries are successful.

also connection from a routine is start from the `prepareBatch` function call.

can check the active connections in clickhouse by `SELECT * FROM system.metrics WHERE metric LIKE '%Connection'`

```
package main

import (
	"context"
	"fmt"
	"net/url"
	"sync"
	"time"

	"github.com/ClickHouse/clickhouse-go/v2"
)

func newClickhouseClient() (clickhouse.Conn, error) {
	// use empty database to create database
	ctx := context.Background()
	dsnURL, err := url.Parse("http://localhost:9000")
	if err != nil {
		return nil, err
	}

	options := &clickhouse.Options{
		Addr:         []string{dsnURL.Host},
		MaxOpenConns: 20,
	}

	if dsnURL.Query().Get("username") != "" {
		auth := clickhouse.Auth{
			Username: dsnURL.Query().Get("username"),
			Password: dsnURL.Query().Get("password"),
		}
		options.Auth = auth
	}
	db, err := clickhouse.Open(options)
	if err != nil {
		return nil, err
	}

	if err := db.Ping(ctx); err != nil {
		return nil, err
	}
	return db, nil
}

func mygo(wg *sync.WaitGroup, db clickhouse.Conn, i int) {
	defer wg.Done()
	retry := 5
	fmt.Printf("starting %d\n", i)
	for x := 0; x < retry; x++ {
		st := `
		INSERT INTO example VALUES (
			?, ?, ?, now()
		)
		`
		//This is where a new connection is created.
		statement, err := db.PrepareBatch(context.TODO(), st)
		if err != nil {
			if x == retry-1 {
				fmt.Printf("failed to ingest for %d , stopping .....\n", i)
				panic(err)
			} else {
				fmt.Printf("failed to ingest for %d , retrying ..... \n", i)
				time.Sleep((time.Duration((x+1)*2) * time.Second))
				continue
			}
		}
		defer func() {
			_ = statement.Abort()
		}()
		for j := 0; j < 10; j++ {
[...]

PR #228, opened 2023-11-15
fix: proper defer statement for log exporter

nityanandagohain:
The defer function was not executed properly in the previous implementation, because of which connections were kept on hold.

Now even if I use maxOpenConnection = numConsumers - 1, once it fails to acquire connection it will retry and it is eventually successful.

Fixes https://github.com/SigNoz/signoz-otel-collector/issues/222


makeavish:
Can we use https://github.com/ClickHouse/clickhouse-go#preparebatch-options?

nityanandagohain:
Thanks have added that @makeavish

nityanandagohain:
After adding withReleaseConnection, it works without abort statement as well, but keeping it as there is no harm. cc @makeavish

srikanthccv (approved):

makeavish (approved):

PR #240, opened 2023-12-09
Write to new v4 table with nearest hour unix precision

srikanthccv:
## What we do today

We maintain two tables in the metrics database

1. samples
2. time_series

The `samples` table contains the metric value and timestamp it was collected at. The `time_series` table contains the metric_name + tags (referred to as timeseries) and the first instance of the timestamp at which the new series was created.

### Two tables

One way to model the schema is to have a single table with all the data.

- value
- timestamp
- tags (either map or pairwise array or json encoded string)

The speed of the query is dependent on the amount of data it has to read. Reading labels from map/array/json encoded string is slow since it has to read the full labels and unpack it (map is just an array underneath). So the more rows you have the slower the query will be. One important characteristic of the timeseries data is that it doesn't change very often. We can have one set of labels for a metric that reports for days, weeks or even months. So we maintain a separate table for timeseries and samples. This improves the speed of the query since we only have to read the labels once and then join them with the samples table.

### Conditional insert to time series table

If there is no change in the labels then we don't need to insert a new row in the time series table. We can just use the existing row. The collector maintains a lookup table of timeseries and their ids. If the timeseries is not in the lookup table then it inserts a new row in the time series table and updates the lookup table. If the timeseries is in the lookup table then the row is not inserted in the time series table.

## Issues with the current implementation

1. The stale timeseries are not deleted from the time series table. 
    - The tag key and value suggestion in the UI will show stale tags.
    - The growing size of the table with stale time series will slow down the query.

2. We don't know the last time a timeseries was seen (commonly known as active time series). This is important to know if we want to delete the stale timeseries.

## This PR
Insert the time series with the hour precision regardless of the prior existence of the same time series. We want to test
- how effective the ReplacingMergeTree engine is and how much space it saves. 
- how does this compare to the current implementation

This is disabled by default. We will do this exercise for a selected set of users in production and asses its effectiveness.

### DIFF
--- a/migrationmanager/migrators/metrics/migrations/000003_ts_v4.up.sql
+++ b/migrationmanager/migrators/metrics/migrations/000003_ts_v4.up.sql
@@ -1,0 +1,19 @@
+CREATE TABLE IF NOT EXISTS signoz_metrics.time_series_v4 ON CLUSTER {{.SIGNOZ_CLUSTER}} (
+    env LowCardinality(String) DEFAULT 'default',
+    temporality LowCardinality(String) DEFAULT 'Unspecified',
+    metric_name LowCardinality(String),
+    description LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
+    unit LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
+    type LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
+    is_monotonic Bool DEFAULT false CODEC(ZSTD(1)),
+    fingerprint UInt64 CODEC(Delta, ZSTD),
+    unix_milli Int64 CODEC(Delta, ZSTD),
+    labels String CODEC(ZSTD(5)),
+    INDEX idx_labels labels TYPE ngrambf_v1(4, 1024, 3, 0) GRANULARITY 1
+)
+ENGINE = ReplacingMergeTree
+        PARTITION BY toDate(unix_milli / 1000)
+        ORDER BY (env, temporality, metric_name, fingerprint, unix_milli)
+        TTL toDateTime(unix_milli/1000) + INTERVAL 2592000 SECOND DELETE;
+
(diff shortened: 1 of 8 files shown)


migrationmanager/migrators/metrics/migrations/000003_ts_v4.up.sql
```diff
@@ -0,0 +1,19 @@
...
+    type LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
+    is_monotonic Bool DEFAULT false CODEC(ZSTD(1)),
+    fingerprint UInt64 CODEC(Delta, ZSTD),
+    unix_milli Int64 CODEC(Delta, ZSTD),
+    labels String CODEC(ZSTD(5)),
+    INDEX idx_labels labels TYPE ngrambf_v1(4, 1024, 3, 0) GRANULARITY 1
+)
+ENGINE = ReplacingMergeTree
+        PARTITION BY toDate(unix_milli / 1000)
+        ORDER BY (env, temporality, metric_name, fingerprint, unix_milli)
```
ankitnayan, line 16: what is the logic for de-duplication by ReplacingMergeTree? If it's sorting key, then `unix_milli` being part of the order by clause will not lead to deduplication?

migrationmanager/migrators/metrics/migrations/000003_ts_v4.up.sql
```diff
@@ -0,0 +1,19 @@
...
+        TTL toDateTime(unix_milli/1000) + INTERVAL 2592000 SECOND DELETE;
```
ankitnayan, line 17: `SETTINGS ttl_only_drop_parts=1;` not needed?

migrationmanager/migrators/metrics/migrations/000003_ts_v4.up.sql
ankitnayan: I see this `unix_milli` is hourly timestamp, right? Then above will should be fine

ankitnayan (approved):

ankitnayan:
We would be writing to this table at every batch insert then, right?

migrationmanager/migrators/metrics/migrations/000003_ts_v4.up.sql
srikanthccv: This table is already better off wrt TTL as the rows would have hourly precision and the table size should be in a reasonable range, unlike the other tables. I will add `ttl_only_drop_parts`.

srikanthccv:
>We would be writing to this table at every batch insert then, right?

Yes

PR #242, opened 2023-12-11
fix: opamp client: ignore non-running collector in ensureRunning when reloading collector config

raj-k-singh:
Right now, if ensureRunning happens to catch a stopped collector svc while a config reload is in progress, it triggers process level shutdown.

Fixes https://github.com/SigNoz/signoz/issues/4181


srikanthccv (approved):
Good catch

PR #243, opened 2023-12-13
readme added

nityanandagohain:
Added note on how to create debug tag


prashant-shahi:
@nityanandagohain  I don't think this should be part of README but rather `publish.md` or an internal documentation.

PR #244, opened 2023-12-13
chore: capture processing times for spanmetrics, tracesexporter, kafk…

dhawal1248:
…aexporter

### DIFF
--- a/exporter/clickhousetracesexporter/clickhouse_factory.go
+++ b/exporter/clickhousetracesexporter/clickhouse_factory.go
@@ -44,10 +44,12 @@ type Writer interface {
 type writerMaker func(WriterOptions) (Writer, error)
 
 var (
-	writeLatencyMillis = stats.Int64("exporter_db_write_latency", "Time taken (in millis) for exporter to write batch", "ms")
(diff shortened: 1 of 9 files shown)


exporter/clickhousetracesexporter/clickhouse_factory.go
```diff
@@ -60,8 +61,17 @@ func ClickHouseNewFactory(migrations string, datasource string, dockerMultiNodeC
 		TagKeys:     []tag.Key{exporterKey, tableKey},
 		Aggregation: writeLatencyDistribution,
 	}
+	processingTimeDistribution := view.Distribution(100, 200, 300, 400, 500, 600, 700, 800, 900, 1000, 1100, 1200, 1300, 1400, 1500, 1750, 2000, 4000, 8000, 16000, 32000, 64000, 128000, 256000, 512000)
```
srikanthccv, line 64: What's the rationale for this bucket layout?

dhawal1248: i wanted latencies in the range 0-1.5s to be as accurate as possible so used this
this is just for debugging..it is not meant to be released

srikanthccv:
If a PR isn't ready just keep it in draft state. What's the point of marking it ready and requesting reviews when it's not meant to be reviewed?

PR #255, opened 2023-12-27
fix: replace ch library with fork

nityanandagohain:
Added a nil pointer check in the library for fixing this issue.

Also created a PR in the upstream repo https://github.com/ClickHouse/clickhouse-go/issues/1167

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -524,6 +524,8 @@ require (
 )
 
 replace (
+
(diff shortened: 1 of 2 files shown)


makeavish (approved):
We need to switch back to official build after latest release.
Hope you sanity tested ingestion

dhawal1248 (approved):

PR #256, opened 2023-12-27
signozkafkareceiver: expose sarama consumer config

dhawal1248:
This PR exposes some of the sarama library's fields like max processing time, max channel size, etc
This would enable us to override these values when required

### DIFF
--- a/receiver/signozkafkareceiver/config.go
+++ b/receiver/signozkafkareceiver/config.go
@@ -32,6 +32,17 @@ type MessageMarking struct {
 	OnError bool `mapstructure:"on_error"`
 }
 
+type SaramaConsumerConfig struct {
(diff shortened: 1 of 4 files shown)


nityanandagohain:
So this PR makes it configurable, but can you also add overall default values, if config is not present?

like some higher timeouts than what's present.

dhawal1248:
> So this PR makes it configurable, but can you also add overall default values, if config is not present?
> 
> like some higher timeouts than what's present.

I was thinking we can rely on the underlying sarama lib for defaults
for deployments that might need these configurations, we can override them

nityanandagohain (approved):
Please create a PR on the charts repos to make the max_processing_time similar to the value of timeout in the exporters

PR #259, opened 2023-12-27
signozkafkareceiver: record processing times

dhawal1248:


nityanandagohain (approved):
LGTM! But do try to add small descriptions to your PR's so that people can get an overview without going into the code.

In this case I already had context but it will help someone else looking into the PR.

PR #264, opened 2024-01-04
Enable WriteTSToV4 for everyone by default

srikanthccv:
See https://github.com/SigNoz/engineering-pod/issues/1059#issuecomment-1874839009


srikanthccv:
I could get rid of the setting alltogether but it's handy and I would like to keep it around for some time and remove it later.

nityanandagohain (approved):

Issue #252, opened 2023-12-21
Past top level operations should be expired


srikanthccv:
As the code changes over time, a top-level operation in the past doesn't necessarily be the top-level operation now as well. The top-level operations table should have the TTL.

PR #266, opened 2024-01-08
Add TTL to top_level_operations table

srikanthccv:
Fixes https://github.com/SigNoz/signoz/issues/4265
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/252
Currently, the `top_level_operations` doesn't have any table. This leads to a couple of issues

- The (test) services are kept in db forever.
- Incorrect top-level operations are kept forever without any cleanup. Assume some operation was identified as top-level operation due to incorrect/broken instrumentation.

### DIFF
--- a/migrationmanager/migrators/traces/migrations/000025_add_time_column_ttl_top_level.up.sql
+++ b/migrationmanager/migrators/traces/migrations/000025_add_time_column_ttl_top_level.up.sql
@@ -1,0 +1,5 @@
+ALTER TABLE signoz_traces.top_level_operations ON CLUSTER {{.SIGNOZ_CLUSTER}} 
+    ADD COLUMN IF NOT EXISTS time DateTime DEFAULT now() CODEC(ZSTD(1));
+
+ALTER TABLE signoz_traces.top_level_operations ON CLUSTER {{.SIGNOZ_CLUSTER}} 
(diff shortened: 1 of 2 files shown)


migrationmanager/migrators/traces/migrations/000025_add_time_column_ttl_top_level.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE signoz_traces.top_level_operations ON CLUSTER {{.SIGNOZ_CLUSTER}} 
+    ADD COLUMN IF NOT EXISTS time DateTime DEFAULT now() CODEC(ZSTD(1));
+
```
nityanandagohain, line 3: are we not considering adding it to the distributed table as well ?

srikanthccv: I thought about it but I didn't see the need to do it because we just want old ones to get TTL'd on the actual tables but no use of that column in the distributed table. I can add it for consistency.

nityanandagohain: yeah that is fine. don't see an issue

nityanandagohain (approved):

PR #218, opened 2023-11-02
fix: upgrade clickhouse client and set maxopenconnections

nityanandagohain:
It fixes this issues i.e

1) **When there is data being filled to queue with num consumers >1 then it panics and shut down**.
  <details>
      <summary>stacktrace</summary>
  ```
  2023-11-02T14:04:31.143+0530	error	exporterhelper/queued_retry.go:307	Dropping data because sending_queue is full. Try increasing queue_size.	{"kind": "exporter", "data_type": "logs", "name": "clickhouselogsexporter", "dropped_items": 512}
  go.opentelemetry.io/collector/exporter/exporterhelper.(*queuedRetrySender).send
	  /Users/nityanandagohain/go/pkg/mod/go.opentelemetry.io/collector/exporter@v0.79.0/exporterhelper/queued_retry.go:307
  go.opentelemetry.io/collector/exporter/exporterhelper.NewLogsExporter.func2
	  /Users/nityanandagohain/go/pkg/mod/go.opentelemetry.io/collector/exporter@v0.79.0/exporterhelper/logs.go:104
  go.opentelemetry.io/collector/consumer.ConsumeLogsFunc.ConsumeLogs
	  /Users/nityanandagohain/go/pkg/mod/go.opentelemetry.io/collector/consumer@v0.79.0/logs.go:25
  go.opentelemetry.io/collector/consumer.ConsumeLogsFunc.ConsumeLogs
	  /Users/nityanandagohain/go/pkg/mod/go.opentelemetry.io/collector/consumer@v0.79.0/logs.go:25
  go.opentelemetry.io/collector/receiver/otlpreceiver/internal/logs.(*Receiver).Export
	  /Users/nityanandagohain/go/pkg/mod/go.opentelemetry.io/collector/receiver/otlpreceiver@v0.79.0/internal/logs/otlp.go:40
  go.opentelemetry.io/collector/pdata/plog/plogotlp.rawLogsServer.Export
	  /Users/nityanandagohain/go/pkg/mod/go.opentelemetry.io/collector/pdata@v1.0.0-rcv0012/plog/plogotlp/grpc.go:82
  go.opentelemetry.io/collector/pdata/internal/data/protogen/collector/logs/v1._LogsService_Export_Handler.func1
	  /Users/nityanandagohain/go/pkg/mod/go.opentelemetry.io/collector/pdata@v1.0.0-rcv0012/internal/data/protogen/collector/logs/v1/logs_service.pb.go:311
[...]

PR #267, opened 2024-01-08
Add time series v4 tables with 6hr, 1day precision

srikanthccv:
This change adds two mvs to the time_series_v4 table, one with 6hrs and the other with 24hrs rounded unix milli. This essentially means we will keep one record for every 6hrs and 24hrs even if the same time series appears again. This helps in reducing the amount of data we need to read when querying. If you need more context https://github.com/SigNoz/engineering-pod/issues/1059#issuecomment-1880151734.

While at it, I also added the release conn and used the consumer settings for conn management. The shard key should be the same for two tables so I added a new sample that has the same shard key as the time series table.


exporter/clickhousemetricsexporter/exporter.go
```diff
@@ -76,13 +76,15 @@ func NewPrwExporter(cfg *Config, set exporter.CreateSettings) (*PrwExporter, err
...
+	maxIdleConnections := cfg.RemoteWriteQueue.NumConsumers + 1
 
 	userAgentHeader := fmt.Sprintf("%s/%s", strings.ReplaceAll(strings.ToLower(set.BuildInfo.Description), " ", "-"), set.BuildInfo.Version)
 
 	params := &ClickHouseParams{
 		DSN:                  cfg.HTTPClientSettings.Endpoint,
 		DropDatabase:         false,
-		MaxOpenConns:         75,
+		MaxIdleConns:         maxIdleConnections,
+		MaxOpenConns:         maxIdleConnections + 10,
```
ankitnayan, line 87: what is the reasoning for this logic?

exporter/clickhousemetricsexporter/clickhouse.go
```diff
@@ -235,7 +241,7 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest, metr
 			fingerprintToName[f] = make(map[string]string)
 		}
 		fingerprintToName[f][nameLabel] = metricName
-		fingerprintToName[f][env] = env
```
ankitnayan, line 238: Was this a bug earlier as I see map key is same as value `env` ? If not, what changed here?

exporter/clickhousemetricsexporter/exporter.go
srikanthccv: There are 6-7 PrepareBatch writes in the metrics exporter. I added a few more just in case.

exporter/clickhousemetricsexporter/clickhouse.go
srikanthccv: Yes, this was a bug

migrationmanager/migrators/metrics/migrations/000004_ts_v4_high_precision.up.sql
```diff
@@ -0,0 +1,77 @@
...
+    temporality LowCardinality(String) DEFAULT 'Unspecified',
+    metric_name LowCardinality(String),
+    description LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
+    unit LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
+    type LowCardinality(String) DEFAULT '' CODEC(ZSTD(1)),
+    is_monotonic Bool DEFAULT false CODEC(ZSTD(1)),
+    fingerprint UInt64 CODEC(Delta, ZSTD),
+    unix_milli Int64 CODEC(Delta, ZSTD),
+    labels String CODEC(ZSTD(5)),
+    INDEX idx_labels labels TYPE ngrambf_v1(4, 1024, 3, 0) GRANULARITY 1
```
ankitnayan, line 33: check if this is supposed to work as we found out having a single row match in every granule will make the algo scan the granule and skip index will not work.

migrationmanager/migrators/metrics/migrator.go
```diff
@@ -32,6 +32,26 @@ func (m *MetricsMigrator) Migrate(ctx context.Context) error {
 			return fmt.Errorf("failed to remove TTL from table %s.%s, err: %s", database, clickhousemetricsexporter.TIME_SERIES_TABLE, err)
 		}
 	}
+
+	// TODO(srikanthccv): remove this after a while
+	// Why?
+	// The table shard key had incorrectly been set to (env, temporality, metric_name, fingerprint, unix_milli) instead of (env, temporality, metric_name, fingerprint)
```
ankitnayan, line 38: shard key `env, temporality, metric_name, fingerprint` is equivalent to just `fingerprint` as it has env, temporality, metric_name already in it?

ankitnayan:
Do you think any other shard key may be more performant in executing queries? Esp joins or computations done in the individual shards?

ankitnayan:
We also need to make changes to the QS to use these MVs, right?

srikanthccv:
>Do you think any other shard key may be more performant in executing queries? Esp joins or computations done in the individual shards?

We are aiming for an equal distribution of data in each shard. Having fingerprints be part of the shard key proved to uniformly distribute data.

srikanthccv:
>
We also need to make changes to the QS to use these MVs, right?

Yes, none of the v4 tables are used in query service now. Eventually all queries will be migrated to v4 tables.

exporter/clickhousemetricsexporter/clickhouse.go
```diff
@@ -254,7 +260,7 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest, metr
 	ch.timeSeriesRW.Unlock()
 
 	err := func() error {
-		statement, err := ch.conn.PrepareBatch(ctx, fmt.Sprintf("INSERT INTO %s.%s (metric_name, temporality, timestamp_ms, fingerprint, labels, description, unit, type, is_monotonic) VALUES (?, ?, ?, ?, ?, ?, ?, ?)", ch.database, DISTRIBUTED_TIME_SERIES_TABLE))
+		statement, err := ch.conn.PrepareBatch(ctx, fmt.Sprintf("INSERT INTO %s.%s (metric_name, temporality, timestamp_ms, fingerprint, labels, description, unit, type, is_monotonic) VALUES (?, ?, ?, ?, ?, ?, ?, ?)", ch.database, DISTRIBUTED_TIME_SERIES_TABLE), driver.WithReleaseConnection())
```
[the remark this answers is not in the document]
srikanthccv: This is the ideal way to work in `PrepareBatch` with the new version of go client library. This has been added in https://github.com/ClickHouse/clickhouse-go/pull/1062

migrationmanager/migrators/metrics/migrator.go
srikanthccv: Right, they should lead to identical distribution today.

migrationmanager/migrators/metrics/migrations/000004_ts_v4_high_precision.up.sql
srikanthccv: We order by fingerprint which is generated after sorting the labels so it should help. The alternative is to have no skip index which is same as today.

ankitnayan: I would recommend to give this a test in real data to see if it works else it is just extra compute + storage with no perf advantage

exporter/clickhousemetricsexporter/exporter.go
ankitnayan: `WithConnectionRelease` should make use of 1 connection for all these 6-7 batches? https://github.com/ClickHouse/clickhouse-go/blob/eb76971a3050d2c83ae5790162e0d52ba87712f8/README.md#preparebatch-options

srikanthccv: I think you are correct. I initially thought wouldn't be immediately released so it might cause a problem with the next preparebatch. Let me go through the changeset once and check.

ankitnayan: Concurrent connections to clickhouse is going to be pain sometime soon. We should keep a watch.

ankitnayan (approved):

srikanthccv:
@nityanandagohain I forgot details about the issue you attempted to fix (or was it fixed completely) around the number of open conns settings. Do you recall what exactly caused the issue and the fix?

ankitnayan:
> @nityanandagohain I forgot details about the issue you attempted to fix (or was it fixed completely) around the number of open conns settings. Do you recall what exactly caused the issue and the fix?

I think it was this https://github.com/SigNoz/signoz-otel-collector/pull/218

srikanthccv:
The `QueueSettings.NumConsumers + 1` should be enough if the connection is released immediately after `.Send()` of preparebatch as that's what it appears to be doing https://github.com/ClickHouse/clickhouse-go/blob/eb76971a3050d2c83ae5790162e0d52ba87712f8/conn_batch.go#L180-L246. Since these write to tables are not happening in separate go routines we can assume that one connection is enough. I guess we added +5 in the #218 because the default settings use `MaxOpenConns = MaxIdleConns + 5`? Is that correct @nityanandagohain?

srikanthccv:
Keeping `MaxOpenConns` a little higher shouldn't cost unlike `MaxIdleConns` which would be retained in the connection pool. And our `MaxIdleConns` is as many as the number of consumers so that makes sense. I am going to merge this now and add the clickhouse server connection metrics. We might have to configure client info to distinguish the connections from collector vs query-service https://github.com/ClickHouse/clickhouse-go#client-info.

Issue #268, opened 2024-01-09
Move usage table creation to migrations


nityanandagohain:
related https://github.com/SigNoz/signoz/issues/3945

PR #271, opened 2024-01-12
fix: usage migration moved out of code to migrations

nityanandagohain:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/268

### DIFF
--- a/migrationmanager/migrators/logs/migrations/000009_add_usage.down.sql
+++ b/migrationmanager/migrators/logs/migrations/000009_add_usage.down.sql
@@ -1,0 +1,2 @@
+DROP TABLE IF EXISTS signoz_logs.usage on cluster {{ .SIGNOZ_CLUSTER }}
+DROP TABLE IF EXISTS signoz_logs.distributed_usage on cluster {{ .SIGNOZ_CLUSTER }}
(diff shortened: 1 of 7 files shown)


migrationmanager/migrators/logs/migrations/000009_add_usage.down.sql
```diff
@@ -0,0 +1,2 @@
+DROP TABLE IF EXISTS signoz_logs.usage on cluster {{ .SIGNOZ_CLUSTER }}
```
srikanthccv, line 1: Did you test the down? This will probably fail if the individual queries must be terminated by `;` otherwise whole contents will be interpreted as a single query.

migrationmanager/migrators/logs/migrations/000009_add_usage.up.sql
```diff
@@ -0,0 +1,15 @@
+--we don't have timestamp in the order by field as we need to update it continuously.
+CREATE TABLE IF NOT EXISTS signoz_logs.usage ON CLUSTER {{ .SIGNOZ_CLUSTER }} (
+    tenant String,
+    collector_id String,
+    exporter_id String,
+    timestamp DateTime,
+    data String
+) ENGINE MergeTree()
+ORDER BY (tenant, collector_id, exporter_id, timestamp)
```
srikanthccv, line 9: order by has the timestamp by the comment above says otherwise.

migrationmanager/migrators/logs/migrations/000009_add_usage.down.sql
nityanandagohain: Aah right, I tested the up but down I just ran the queries manually, have fixed it :)

migrationmanager/migrators/logs/migrations/000009_add_usage.up.sql
nityanandagohain: Aah that was a comment which was added by mistake due to previous implementations in usage.go and was not removed, I copied it to the migrations without checking, have removed it.

srikanthccv (approved):

PR #273, opened 2024-01-24
signozkafkaexporter: use request id as record key + idempotent producers

dhawal1248:
This PR adds the following changes - 
- use requestId as the record key
parse requestId from client metadata and use it as the kafka record's key - this key will be used to deduplicate data with kafka streams
- idempotent producers 
this PR also enables us to configure [idempotent producers](https://developer.confluent.io/patterns/event-processing/idempotent-writer/) by exposing the sarama library's relevant config

### DIFF
--- a/exporter/signozkafkaexporter/utils.go
+++ b/exporter/signozkafkaexporter/utils.go
@@ -1,8 +1,10 @@
 package signozkafkaexporter
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/Shopify/sarama"
 	"go.opentelemetry.io/collector/client"
 )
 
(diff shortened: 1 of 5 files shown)


exporter/signozkafkaexporter/utils.go
```diff
@@ -15,3 +17,24 @@ func getKafkaTopicFromClientMetadata(md client.Metadata) (string, error) {
...
+}
+
+func setRequestIdAsKey(ctx context.Context, messages []*sarama.ProducerMessage) ([]*sarama.ProducerMessage, error) {
+	requestId, err := getRequestIdFromClientMetadata(client.FromContext(ctx).Metadata)
+	if err != nil {
+		return nil, err
+	}
+	for _, message := range messages {
+		message.Key = sarama.StringEncoder(requestId)
+	}
```
srikanthccv, line 38: Why is it assigning the same identifier to each item here? Would that result in losing the legitimate data? Shouldn't there be sequencing here?

srikanthccv:
The idempotent producer is added as an additional mechanism and our primary way of handling the failure before Kafka is to retry from the ingress with the same request-id which is yet to be implemented.

exporter/signozkafkaexporter/utils.go
srikanthccv: Based on the assumption stated here https://github.com/SigNoz/signoz-otel-collector/pull/273/files#diff-06a5e74f9ab7315cfd07374023d50c6ccf75d45afbf188f0047c76dfaf4ce968R52 we would lose data. We should sequence them or if the assumption is not correct, then update the comment and panic when the size is not 1.

PR #285, opened 2024-02-06
feat: drop logs older than TTL

### LINKED ISSUE #272: Discard logs older than the retention period
If an ingested log is older than the retention period skip it from getting ingested.

nityanandagohain:
Fixes #272 

## Summary by CodeRabbit

- **New Features**
    - Improved log management by adding methods to retrieve logs TTL and remove old logs based on a timestamp.
    - Enhanced data pushing to Clickhouse with retry logic and TTL handling.

### DIFF
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -20,6 +20,8 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


exporter/clickhouselogsexporter/exporter.go
```diff
@@ -117,10 +119,88 @@ func (e *clickhouseLogsExporter) Shutdown(_ context.Context) error {
...
+	if err != nil {
+		return delTTL, err
+	}
+	if len(dbResp) == 0 {
+		return delTTL, fmt.Errorf("ttl not found")
+	}
+
+	deleteTTLExp := regexp.MustCompile(`toIntervalSecond\(([0-9]*)\)`)
+
+	m := deleteTTLExp.FindStringSubmatch(dbResp[0].EngineFull)
```
srikanthccv, line 141: Can you confirm if the `[0]` will always be the main TTL of the table?

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -117,10 +119,88 @@ func (e *clickhouseLogsExporter) Shutdown(_ context.Context) error {
...
+	acceptedDateTime := time.Now().Add(-(time.Duration(ttL) * time.Second))
+
+	removeLog := func(log plog.LogRecord) bool {
+		t := log.Timestamp().AsTime()
+		return t.Unix() < acceptedDateTime.Unix()
+	}
+	for i := 0; i < ld.ResourceLogs().Len(); i++ {
+		logs := ld.ResourceLogs().At(i)
+		for j := 0; j < logs.ScopeLogs().Len(); j++ {
+			logs.ScopeLogs().At(j).LogRecords().RemoveIf(removeLog)
```
srikanthccv, line 170: I forgot what is the default state. Is there any chance of panic in our setup?

srikanthccv (approved):

exporter/clickhouselogsexporter/exporter.go
nityanandagohain: Didn't get you here

exporter/clickhouselogsexporter/exporter.go
nityanandagohain: Yes, even if we add cold storange the query will return something like this 

```
MergeTree PARTITION BY toDate(timestamp / 1000000000) ORDER BY (timestamp, id) 
TTL toDateTime(timestamp / 1000000000) + toIntervalSecond(1296000), 
toDateTime(timestamp / 1000000000) + toIntervalSecond(72000) TO VOLUME 's3' SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1, storage_policy = 'tiered'
```

And FindSubstringmath returns the `leftmost match` , I have used the same logic present when we load our TTL in query service

exporter/clickhouselogsexporter/exporter.go
srikanthccv: `RemoveIf` mutates the own state. If it's state is immutable then it panics. You can check the implementation of `RemoveIf`. I just checked this shouldn't have any issue in our case.

Issue #283, opened 2024-02-06
The initial value service.instance.id is overwritten and displayed as an hex encoded string


cedstrom:
[See signoz #4388](https://github.com/SigNoz/signoz/issues/4338)

A change in the otel collector has caused the instance id to show up as a hex string rather than as the actual name of the service.

Potential fix is also included: https://github.com/sharpSteff/signoz-otel-collector/commit/e8005a542c67a16342ab4664f2c2876f0f8565fb

cedstrom:
@srikanthccv, the regression seems to have been isolated to your commit: https://github.com/SigNoz/signoz-otel-collector/commit/d1dc659eae9eddf93c88bf567d4346f105601789.  Can you take a look?

srikanthccv:
For the span metrics, the collector is the source that generates the metrics. The idea behind the `service.instance.id` is to introduce uniqueness and produce a unique time series from each metric generator. If this were not the case, two replicas of the signoz-otel-collector would produce an identical time series which would lead to incorrect results. I guess the ideal fix is to not replace the original `service.instance.id` but continue to use `signoz.collector.id` to guarantee single-write metric stream https://opentelemetry.io/docs/specs/otel/metrics/data-model/#single-writer

PR #289, opened 2024-02-13
Do not replace service.instance.id from SDK

srikanthccv:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/283
Fixes https://github.com/SigNoz/signoz/issues/4338

In the past, we created our own ID `signoz.collector.id` for the uniqueness guarantees and then moved to use the `service.instance.id` from the collector `TelemetrySettings`. This conflicts with the `service.instance.id` coming from the Client SDK with the signoz-otel-collector. Users want to see their instance instead of the SigNoz deployment under the `service.instance.id` attribute. This is a reasonable ask. We will honour their settings and maintain our own attribute `signoz.collector.id` as we did earlier. 

## Summary by CodeRabbit

- **Refactor**
	- Updated the handling of instance IDs in metrics aggregation for improved identification and tracking.

### DIFF
--- a/processor/signozspanmetricsprocessor/processor.go
+++ b/processor/signozspanmetricsprocessor/processor.go
@@ -33,7 +33,6 @@ import (
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
-	semconv "go.opentelemetry.io/collector/semconv/v1.13.0"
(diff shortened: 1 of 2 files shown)


nityanandagohain (changes requested):
Please update the test as well https://github.com/SigNoz/signoz-otel-collector/blob/b531173042168c448b59dc6ddc4a1226549c9472/processor/signozspanmetricsprocessor/processor_test.go#L829

nityanandagohain (approved):
LGTM!

Please make sure to add this in release notes.

PR #294, opened 2024-02-27
Use clickhouse.ParseDSN for clickhouse URL parsing

srikanthccv:


srikanthccv:
Easy one, please take a look. Uses the right connection string parsing. It enables secure connection b/w the collector and ClickHouse. All additional settings can also be provided.

nityanandagohain (approved):

PR #304, opened 2024-03-21
(feat) return a default topic when tenant_id is not present

grandwizard28:
#### Features

- Return a default topic name when `signoz_tenant_id` is missing in the metadata

#### Refactors

- Change name of the function `getKafkaTopicPrefixFromClientMetadata` to reflect what it is actually doing.

### DIFF
--- a/exporter/signozkafkaexporter/utils.go
+++ b/exporter/signozkafkaexporter/utils.go
@@ -5,16 +5,16 @@ import (
 )
 
 const (
-	DefaultKafkaTopic = "default"
+	DefaultKafkaTopicPrefix = "default"
 )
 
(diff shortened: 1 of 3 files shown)


exporter/signozkafkaexporter/utils.go
```diff
@@ -1,17 +1,20 @@
 package signozkafkaexporter
 
 import (
-	"fmt"
-
 	"go.opentelemetry.io/collector/client"
 )
 
+const (
+	DefaultKafkaTopic = "default"
```
prashant-shahi, line 8:
```suggestion
	DefaultKafkaTopicPrefix = "default"
```

prashant-shahi (approved):

srikanthccv (approved):

PR #309, opened 2024-04-01
feat: change body index to ngram

nityanandagohain:
migration to add ngram index to body.

fixes https://github.com/SigNoz/signoz/issues/4259


srikanthccv:
A short length ngram also means there is a scope of false positives in many contexts. Imagine the IP address being part of the body and the IP range is 10.8.x.y - for such a dataset all granules may match because the prefix matches. This can manifest in several forms and becomes a problem as the more diverse users we onboard. Our benchmarks are only as good as our test data. 

This fix will help some users and we should roll it out but I'd also suggest continuing to look into more options if possible.

nityanandagohain:
I think it depends more on the cardinality rather than perfixes, if your data is higly cardinal the index will be able to benifit the query.

In the above on if the ip adress is highly cardinal it will be able to skip granules easily. because even the last ngram will result in a unique hash.

srikanthccv (approved):
>Another good candidate for a skip index is for high cardinality expressions where any one value is relatively sparse in the data. One example might be an observability platform that tracks error codes in API requests. Certain error codes, while rare in the data, might be particularly important for searches. A set skip index on the error_code column would allow bypassing the vast majority of blocks that don't contain errors and therefore significantly improve error focused queries.


> Finally, the key best practice is to test, test, test. Again, unlike b-tree secondary indexes or inverted indexes for searching documents, data skipping index behavior is not easily predictable. Adding them to a table incurs a meangingful cost both on data ingest and on queries that for any number of reasons don't benefit from the index. They should always be tested on real world type of data, and testing should include variations of the type, granularity size and other parameters. Testing will often reveal patterns and pitfalls that aren't obvious from thought experiments alone.


https://clickhouse.com/docs/en/optimize/skipping-indexes#skip-index-settings


I was mainly interested in the false positives regardless of the cardinality. I don't have any data to prove my point but what I recall from the past experiments was that collision was high and led to more false positives. 


Let's ship this and see how it performs in varying real world datasets.

nityanandagohain:
@srikanthccv I have updated only the value of index to 60KB from 50KB, please have a look once.

srikanthccv (approved):

PR #313, opened 2024-04-12
feat(admin): add admin apis for collector

grandwizard28:
#### Description
The collector has been converted to a cobra based cli with the following commands:
```
Available Commands:
  api         Starts an api to interact with configuration
  migrate     Manage database schema migrations
  run         Run the otel collector
```

A new component api has been added with an optional storage backend `postgres`. The api (or admin api) will be responsible for storing and serving configuration.

#### Features

- Create an admin api for configuration
- Add postgres schemas using atlas
- Refactor cmd to a cobra based cli
- Add global dependency injection for collector
- Add `keyauth` extension for authentication by postgres
- Add `limiter` processor for rate limiting by size and count.

#### Usage
A very basic config file:
```
receivers:
  otlp:
    protocols:
      grpc:
      http:
        auth:
          authenticator: keyauth
processors:
  batch:
    send_batch_size: 1000
    timeout: 10s
  limiter:
    policy: redis/postgres
extensions:
  keyauth:
    headers:
      - signoz-access-token
      - X-Signoz-Token
exporters:
  logging:
    verbosity: detailed
service:
  extensions: [keyauth]
  pipelines:
    logs:
      receivers: [otlp]
      processors: [limiter, batch]
      exporters: [logging]
```


pkg/processor/limiter/processor.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (lp *limiterProcessor) consume(ctx context.Context, signal entity.Signal, limitValue entity.LimitValue) error {
+	// Retrieve credential id from ctx
+	keyId, ok := entity.KeyIdFromCtx(ctx)
+	if !ok {
+		return nil
+	}
+
+	// Retrieve the limit from the backing store
+	limit, err := lp.storage.DAO.Limits().SelectByKeyAndSignal(ctx, keyId, signal)
```
srikanthccv, line 90: The limits do not change frequently. It will be fine to fetch the limits once every `fetchInterval`.

pkg/processor/limiter/processor.go
```diff
@@ -0,0 +1,146 @@
...
+		limitMetrics = append(limitMetrics, entity.NewLimitMetric(
+			period,
+			limitValue,
+			signal,
+			entity.PeriodAt(currTime)[period],
+			keyId,
+		))
+	}
+
+	err = lp.storage.DAO.LimitMetrics().Insert(ctx, limitMetrics)
```
srikanthccv, line 140: The point we were trying to make during the discussion was that the processor would maintain a temporary local state of aggregation for a minute or so and flush them in intervals. The current limits also do not need to be fetched for each consume request.

pkg/entity/period.go
```diff
@@ -0,0 +1,120 @@
...
+// It implements the driver.Valuer and sql.Scanner interfaces for
+// working with databases,and the json.Marshaler and json.Unmarshaler
+// interfaces for JSON serialization and deserialization.
+type Period struct {
+	name string
+}
+
+var (
+	PeriodSecond = Period{name: "second"}
+	PeriodMinute = Period{name: "minute"}
```
srikanthccv, line 22: I don't see any practical use case where users want to limit the count/size every second/minute basis.

pkg/processor/limiter/factory.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/processor"
+	"go.opentelemetry.io/collector/processor/processorhelper"
+)
+
+func NewFactory() processor.Factory {
+	return processor.NewFactory(
+		component.Type("limiter"),
```
srikanthccv, line 14: Let's prepend signoz for our custom components.

srikanthccv (commented):
Some initial review. I mainly looked at the processor/keyauth logic

cmd/run.go
```diff
@@ -0,0 +1,137 @@
...
+	)
+	logger.Infoctx(ctx, "initialized storage")
+
+	cache := cache.NewCache(
+		cache.WithHost(cacheConfig.host),
+		cache.WithPort(cacheConfig.port),
+	)
+
+	//Inject the global env here
+	env.NewG(env.WithStorage(storage), env.WithCache(cache))
```
srikanthccv, line 98: Let's not introduce a global state. It is frowned upon in go and I agree with that. https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html, https://web.archive.org/web/20190508022304/https://twitter.com/davecheney/status/871939730761547776. There is a reason why `golangci-lint` (we don't use) has the linter checks for `gochecknoglobals`.

cmd/api.go
```diff
@@ -0,0 +1,78 @@
...
+				cmd.Context(),
+				logger,
+				storageConfig,
+				adminHttpConfig,
+			)
+		},
+	}
+
+	adminHttpConfig.registerFlags(cmd)
+	storageConfig.registerFlags(cmd)
```
srikanthccv, line 37: What do you think about adding a config file instead for API options? That would remain consistent with the rest of the collector components and make adding more options for sub-components separate from the high-level cli flags.

cmd/run.go
```diff
@@ -0,0 +1,137 @@
...
+				storageConfig,
+				collectorConfig,
+				cacheConfig,
+			)
+		},
+	}
+
+	collectorConfig.registerFlags(cmd)
+	storageConfig.registerFlags(cmd)
+	cacheConfig.registerFlags(cmd)
```
srikanthccv, line 46: I see these are used in extension/processor components. They should rather be part of the component config section instead of here.

pkg/storage/strategies/strategy.go
```diff
@@ -0,0 +1,57 @@
...
+//		Set(string) error
+//		Type() string
+//	}
+type Strategy struct {
+	name string
+}
+
+var (
+	Postgres = Strategy{name: "postgres"}
+	Off      = Strategy{name: "off"}
```
srikanthccv, line 20: Redis should be another strategy here?

cmd/run.go
```diff
@@ -0,0 +1,127 @@
...
+func runCollector(
+	ctx context.Context,
+	logger log.Logger,
+	storageConfig storageConfig,
+	collectorConfig collectorConfig,
+) error {
+	// Copy files
+	if collectorConfig.managerConfig != "" {
+		// Copy the collector config to the copy path ??
+		// Why?
```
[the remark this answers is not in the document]
grandwizard28: Will update this in the comments!

pkg/processor/limiter/processor.go
grandwizard28: Users can change limits from the UI. In addition, some users have requested a feature to stop ingestion for a specified period of time for which they can set the limits to 0.

Caching calls to save DB queries can be picked up later as a general theme to improve performance.

cmd/run.go
grandwizard28: Agreed it's not the best practice. 

The entire idea is to open a single db connection and have it used across all extensions/recievers/processors. Otherwise, the keyauth and the limiter extension both would need to open a DB connection separately which is suboptimal.

cmd/api.go
grandwizard28: I was initially thinking of the same but managing credentials is a pain. The precedence defined currently is:
- flag
- env

We can add a config file precedence to this but I still would like to keep the flags and env.
- flag
- env
- config file

pkg/storage/strategies/strategy.go
grandwizard28: This is strategies related to a backing store. Redis and lru, for example could be cache strategies.

pkg/processor/limiter/processor.go
grandwizard28: The aggregation has to be done on `period, signal, period_at, key_id`. Unless we use a consistent hash load balancer in front to ensure all requests with the same `key_id` hit the same instance, we would just be delaying the insertion of all independent metrics.

cmd/run.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if err := collector.Start(ctx); err != nil {
+		logger.Errorctx(ctx, "failed to start the collector", err)
+		return err
+	}
+
+	wait(ctx, logger, collector.Error())
+	// TODO: Move this cancel logic inside collector.Shutdown()
+	// ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	// defer cancel()
```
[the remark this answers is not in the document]
grandwizard28: I was thinking that we can release this independently of the existing codebase. Once stable, we can resolve this TODO comment. I want to touch the existing code as less as possible.

pkg/processor/limiter/processor.go
srikanthccv: I know they can change the limits from the UI and can be set to any number. What I am trying to say is that they are not frequent. We pick a sensible fetchInterval and only fetch the details of that interval instead of fetching it for each consume call. Say we choose the `fetchInterval` as 1 minute, now the maximum delay in applying the limits is a maximum of one minute which is fine in all scenarios.

pkg/processor/limiter/processor.go
srikanthccv: It doesn't have to be consistent. The trade-off is that the local state is behind the actual value and allows ingesting a little bit more than actually permitted. And with a reasonable flush interval, that over commit of data won't make much difference.

cmd/api.go
srikanthccv: Why is managing credentials pain?

pkg/storage/strategies/strategy.go
srikanthccv: I am a little confused. My understanding is that there is a store where we maintain the counters of the received data. This store can be redis, postgres or other impl. What is this referred to as? I know there is a postgres store where we store the limits from UI and tenant details

cmd/run.go
srikanthccv: How much does it affect if each component opens its own connection? Is that a problem to worry about? As you can it already, it's quite hard to follow how do components work. And once we introduce it into the codebase it's only going to get (ab)used more.

pkg/storage/strategies/strategy.go
grandwizard28: This is the backing store for all configurations, keys and tenants which will always be postgres.

The storage of counters can be postgres or redis. Depends on the policy selected in the `signozlimiter` processor.

cmd/run.go
grandwizard28: In this case alone,
One connection pool for keyauth + One connection pool for limiter + Separate connection pool if other extensions/processors decide to use this per **gateway instance**.

Abuse part should be prevented by effective PR reviews!

cmd/api.go
grandwizard28: How do you inject credentials into a container using any orchestration engine?

pkg/processor/limiter/processor.go
grandwizard28: This is applicable to all queries, even the query used in the `keyauth` processor. Can pick it up as theme later altogether to improve performance.

pkg/processor/limiter/processor.go
srikanthccv: I don't follow. I am saying there is no need to have a consistent hash load balancer because the count/sizes can lag for the duration of the flush interval. The component maintains a local aggregation state per each unique period, signal, key_id etc.. And update the count/size once every flush interval.

For example, take key `k1` and four collectors `c1-4`. For the sake of argument, say the `consume` was called 10000 times for with `k1` and counts are `c1-2000, c2-3000, c3-4000, c5-1000`. The collector does the local aggregation for a minute and at the end of the minute interval it updates the counts in the persistent store.

cmd/api.go
srikanthccv: The config file will reference an env like `password: ${PASSWORD}` and the env gets injected before the config gets applied?

cmd/run.go
srikanthccv: I think one additional connection is not an issue per gateway instance because the number of gateway instances is relatively small. Once a pattern is introduced to the codebase there is no going back. I do not have a strong opinion but at this point, I see keeping the config and connection separate for components as a better option and revisit when it becomes a problem.

pkg/storage/strategies/strategy.go
srikanthccv: Thanks, I understand it now.

PR #336, opened 2024-06-17
feat: add instrumentation scope support

nityanandagohain:
Otel collector changes for  https://github.com/SigNoz/signoz/issues/2705

### DIFF
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -249,6 +249,7 @@ func (e *clickhouseLogsExporter) pushToClickhouse(ctx context.Context, ld plog.L
 			resources = addTemporaryUnderscoreSupport(resources)
 
 			for j := 0; j < logs.ScopeLogs().Len(); j++ {
+				scope := logs.ScopeLogs().At(j).Scope().Name()
 				rs := logs.ScopeLogs().At(j).LogRecords()
 				for k := 0; k < rs.Len(); k++ {
 					r := rs.At(k)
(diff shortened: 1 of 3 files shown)


srikanthccv:
It also has attributes https://github.com/open-telemetry/opentelemetry-proto/blob/bd7cf55b6d45f3c587d2131b68a7e5a501bdb10c/opentelemetry/proto/common/v1/common.proto#L69-L81; Do we not want to support them for now?

nityanandagohain:
I was planning to just introduce the name first, but I guess it's better to introduce the version as well.

srikanthccv:
I suggest adding the attributes as well so you don't have to come back to it again in future if somebody asks for it.

nityanandagohain:
hmm makes sense, regarding the implementation of it do see any cons with the attributes in a similar way to that of resource attributes. Here will we will name it as instrumentation_scope_string_key, instrumentation_scope_string_value.

srikanthccv:
I don't see any issue.

nityanandagohain:
have added support for name, version and attributes.

srikanthccv (approved):

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -249,6 +249,17 @@ func (e *clickhouseLogsExporter) pushToClickhouse(ctx context.Context, ld plog.L
 			resources = addTemporaryUnderscoreSupport(resources)
 
 			for j := 0; j < logs.ScopeLogs().Len(); j++ {
+				scope := logs.ScopeLogs().At(j).Scope()
+				scopeName := scope.Name()
+				scopeVersion := scope.Version()
+
+				scopeAttributes := attributesToSlice(scope.Attributes(), true)
+
+				err := addAttrsToTagStatement(tagStatement, "instrumentation_scope", resources)
```
srikanthccv, line 258: This should be using `scopeAttributes`

nityanandagohain: Have updated it to `scope` https://github.com/SigNoz/signoz-otel-collector/pull/344

PR #341, opened 2024-06-26
Update migrations for up/down after DROP schema_migrations

srikanthccv:
The main table `signoz_index_v2` dropped some columns which are referenced in the prev migrations which creates a lot of issues because all the old migrations can't be run because the column they use in creating the mv no longer exist and break the migrations. This removes all such past instances so that the migration can be run ever after dropping the schema_migrations tables.


migrationmanager/migrators/traces/migrations/000011_init_durationMV.up.sql
```diff
@@ -35,29 +35,4 @@ CREATE TABLE IF NOT EXISTS signoz_traces.durationSort ON CLUSTER {{.SIGNOZ_CLUST
 PARTITION BY toDate(timestamp)
 ORDER BY (durationNano, timestamp)
 TTL toDateTime(timestamp) + INTERVAL 1296000 SECOND DELETE
-SETTINGS index_granularity = 8192;
```
nityanandagohain, line 38: remove from 11.down as well.

nityanandagohain (approved):
you can just remove `DROP VIEW IF EXISTS signoz_traces.durationSortMV ON CLUSTER {{.SIGNOZ_CLUSTER}};`  from down files to keep it clean and also `DROP VIEW IF EXISTS signoz_traces.dependency_graph_minutes_db_calls_mv_v2 ON CLUSTER {{.SIGNOZ_CLUSTER}};` this

PR #343, opened 2024-07-03
feat: rename instrumentation_scope to scope

nityanandagohain:
part of https://github.com/SigNoz/signoz/issues/2705

Renames `instrumentation_scope, instrumentation_scope_version, instrumentation_scope_attributes_string_{key/val}`

to

`scope_name , scope_version , scope_string_{key,value}`

### DIFF
--- a/migrationmanager/migrators/logs/migrations/000012_rename_instrumentation_scope.up.sql
+++ b/migrationmanager/migrators/logs/migrations/000012_rename_instrumentation_scope.up.sql
@@ -1,0 +1,16 @@
+ALTER TABLE signoz_logs.logs DROP INDEX IF EXISTS instrumentation_scope_idx;
+
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope to scope_name;
+ALTER TABLE signoz_logs.distributed_logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope to scope_name;
+
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_version to scope_version;
+ALTER TABLE signoz_logs.distributed_logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_version to scope_version;
+
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_attributes_string_key to scope_string_key;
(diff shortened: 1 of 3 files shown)


migrationmanager/migrators/logs/migrations/000012_rename_instrumentation_scope.up.sql
```diff
@@ -0,0 +1,16 @@
...
+ALTER TABLE signoz_logs.distributed_logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_version to scope_version;
+
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_attributes_string_key to scope_string_key;
+ALTER TABLE signoz_logs.distributed_logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_attributes_string_key to scope_string_key;
+
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_attributes_string_value to scope_string_value;
+ALTER TABLE signoz_logs.distributed_logs ON CLUSTER {{.SIGNOZ_CLUSTER}} RENAME column IF EXISTS instrumentation_scope_attributes_string_value to scope_string_value;
+
+
+ALTER TABLE signoz_logs.logs ON CLUSTER {{.SIGNOZ_CLUSTER}} ADD INDEX IF NOT EXISTS scope_name_idx (scope_name) TYPE tokenbf_v1(10240, 3, 0) GRANULARITY 4;
```
srikanthccv, line 16: Is this better index for the scope name? The scope name usually contains a dot separated values like `io.signoz.app.user` etc...

nityanandagohain: So the assumtion that lead to use of token bf is that peope will search more for absolute values like

name = io.signoz.app
name in (io.signoz.app, io.signoz.app.user)
name contains "%io.signoz.app%"

i.e alphanumberic characters will most likely be present which will result in formation of tokens and thus usage of index.

srikanthccv: Thanks for the explanation. When does the index get applied? What does the query prepared by query-service look like?

nityanandagohain: it will look like 

`where scope_name = 'io.signoz.app.user'`
`where scope_name in ('io.signoz.app.user', 'io.signoz.app')`
`where scope_name like '%app.user%'`


Now this queries while result will tokens like `'io' , 'signoz' , 'app',  'user'` which will be checked on the index

srikanthccv (approved):

PR #346, opened 2024-07-09
fix: update logs migrations to remove issues

nityanandagohain:
The current migrations for logs gets stuck if the schema migrations table is deleted and the entire migrations are rerun as there are some queries which take a lot of time.

This fix ensures that we can run them again and also correct schema is created when a new instance is created.


migrationmanager/migrators/logs/migrations/000010_body_ngram.up.sql
```diff
@@ -1,2 +1 @@
-ALTER TABLE signoz_logs.logs ON CLUSTER {{ .SIGNOZ_CLUSTER }} DROP INDEX IF EXISTS body_idx ;
-ALTER TABLE signoz_logs.logs ON CLUSTER {{ .SIGNOZ_CLUSTER }} ADD INDEX IF NOT EXISTS body_idx lower(body) TYPE ngrambf_v1(4, 60000, 5, 0) GRANULARITY 1;
\ No newline at end of file
+SELECT 1
```
srikanthccv, line 1: Does this mean those who go from earlier version such as `0.47.0` to new release will not have the body index as `ngrambf_v1 `?

nityanandagohain: Yes this assumes that, 0.48.0 is installed if they are using older version of signoz.

srikanthccv: I think we should at least provide a migration script that triggers these operations separately because if the body index makes a non-trivial difference in #granules read then everyone should have it, including those who do not necessarily on 0.48

nityanandagohain: @srikanthccv I was thinking to write this doc on how to recover if someone is migrating in between this versions, https://signoz-community.slack.com/archives/C01HWQ1R0BC/p1721400860382799 Will that be a good start to this ?

nityanandagohain: @srikanthccv  can we go ahead with merging this ? As I have released the doc https://github.com/SigNoz/signoz-web/pull/527

There will be one more doc which I will work on once a new release with this changes are done. i.e how to migrate from older version directly to latest by skipping `V0.49.1` or `V0.50.0

srikanthccv: Please keep the original command but comment and link to docs.

nityanandagohain: have made the changes @srikanthccv

srikanthccv (approved):

PR #348, opened 2024-07-10
feat: support for new logs schema

nityanandagohain:
New logs schema

fixes https://github.com/SigNoz/signoz/issues/5555

---- testing /running
1) Make sure macro is enabled for running migrations 
```
    <macros>
        <shard>01</shard>
        <replica>example01-01-1</replica>
    </macros>
 ```
 2) Run migrations using `go run cmd/signozschemamigrator/migrate.go --dsn http://localhost:9000 --replication true`

### CODE REVIEW

File: migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql [hunk 1]
```diff
@@ -0,0 +1,121 @@
...
+    INDEX id_minmax id TYPE minmax GRANULARITY 1,
+    INDEX severity_number_idx severity_number TYPE set(25) GRANULARITY 4,
+    INDEX severity_text_idx severity_text TYPE set(25) GRANULARITY 4,
+    INDEX trace_flags_idx trace_flags TYPE bloom_filter GRANULARITY 4,
+    INDEX scope_name_idx scope_name TYPE tokenbf_v1(10240, 3, 0) GRANULARITY 4,
+    INDEX attributes_string_idx_key mapKeys(attributes_string) TYPE tokenbf_v1(1024, 2, 0) GRANULARITY 1,
+    INDEX attributes_string_idx_val mapValues(attributes_string) TYPE ngrambf_v1(4, 5000, 2, 0) GRANULARITY 1,
+    INDEX attributes_int64_idx_key mapKeys(attributes_number) TYPE tokenbf_v1(1024, 2, 0) GRANULARITY 1,
+    INDEX attributes_int64_idx_val mapValues(attributes_number) TYPE bloom_filter GRANULARITY 1,
+    INDEX attributes_bool_idx_key mapKeys(attributes_bool) TYPE tokenbf_v1(1024, 2, 0) GRANULARITY 1
```
nityanandagohain, line 54: no index on resources as we use the resource attribute table for filtering with resources

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -558,6 +699,51 @@ const (
...
+							severity_text,
+							severity_number,
+							body,
+							attributes_string,
+							attributes_number,
+							attributes_bool,
+							resources_string,
+							scope_name,
+							scope_version,
+							scope_string_key,
```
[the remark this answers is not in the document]
nityanandagohain: ahh right.

exporter/clickhouselogsexporter/logsv2/fingerprint.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return result
+}
+
+// TODO(Raj/Nitya): Consider parsing this stuff out from json
+func ResourceHierarchy() *DimensionHierarchyNode {
+	return &DimensionHierarchyNode{
+		labels: []string{
+			"cloud.provider",
+			"source_type",
```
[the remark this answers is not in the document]
nityanandagohain: so `source_type` is for vector logs. It is by default added by vector.

eg:- `k8s`, `file` . So it's kind of top-level for vector

https://signoz.io/docs/logs-management/send-logs/vector-logs-to-signoz/#mapping-for-signoz

exporter/clickhouselogsexporter/logsv2/fingerprint.go
```diff
@@ -0,0 +1,195 @@
...
+
+				subHierachies: []DimensionHierarchyNode{{
+					labels: []string{"cloud.platform"},
+
+					subHierachies: []DimensionHierarchyNode{{
+						labels: []string{
+							"k8s.cluster.name",
+							"k8s.cluster.uid",
+							"aws.ecs.cluster.arn",
+							"gcp.project",
```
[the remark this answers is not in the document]
nityanandagohain: makes sense.

exporter/clickhouselogsexporter/logsv2/hash.go
```diff
@@ -0,0 +1,52 @@
+package logsv2
```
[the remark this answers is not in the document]
nityanandagohain: I don't have an opinion on this, but better to keep it separated for now? maybe we can move to common utils in the future which both the exporters can import.

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -41,15 +42,19 @@ import (
 )
 
 const (
-	DISTRIBUTED_LOGS_TABLE     = "distributed_logs"
-	DISTRIBUTED_TAG_ATTRIBUTES = "distributed_tag_attributes"
+	DISTRIBUTED_LOGS_TABLE                      = "distributed_logs"
+	DISTRIBUTED_TAG_ATTRIBUTES                  = "distributed_tag_attributes"
+	DISTRIBUTED_LOGS_TABLE_V2                   = "distributed_logs_v2"
+	DISTRIBUTED_LOGS_RESOURCE_BUCKET_V2         = "distributed_logs_v2_resource_bucket"
```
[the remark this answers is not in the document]
nityanandagohain: sure, let me get review from others will make this change at end as it will require migration for users who are currently using image built from this

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -240,7 +266,15 @@ func (e *clickhouseLogsExporter) pushToClickhouse(ctx context.Context, ld plog.L
...
 			resources := attributesToSlice(res.Attributes(), true)
 
-			err := addAttrsToTagStatement(tagStatement, "resource", resources)
+			resourcesMap := attributesToMap(res.Attributes(), true)
+
+			serializedRes, err := json.Marshal(res.Attributes().AsRaw())
+			if err != nil {
+				return fmt.Errorf("couldn't serialize log resource JSON: %w", err)
+			}
+			resourceJson := string(serializedRes)
```
srikanthccv, line 275: This is not guaranteed to produce a consistent json string even for the same set of attributes if their order of insertion is not the same in map. What happens if the `ResourceLogs` has multiple entries with same attrs but different order?

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
```diff
@@ -0,0 +1,119 @@
+CREATE TABLE IF NOT EXISTS  signoz_logs.logs_v2_resource_bucket ON CLUSTER {{.SIGNOZ_CLUSTER}}
+(
+    `labels` String CODEC(ZSTD(5)),
+    `fingerprint` String CODEC(ZSTD(1)),
+    `seen_at_ts_bucket_start` Int64 CODEC(Delta(8), ZSTD(1)),
+    INDEX idx_labels lower(labels) TYPE ngrambf_v1(4, 1024, 3, 0) GRANULARITY 1
+)
+ENGINE = ReplacingMergeTree
```
srikanthccv, line 8: Add `{{.SIGNOZ_REPLICATED}}` for all engine statements

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
```diff
@@ -0,0 +1,119 @@
...
+SETTINGS ttl_only_drop_parts = 1, index_granularity = 8192;
+
+
+CREATE TABLE IF NOT EXISTS  signoz_logs.distributed_logs_v2_resource_bucket ON CLUSTER {{.SIGNOZ_CLUSTER}}
+(
+    `labels` String CODEC(ZSTD(5)),
+    `fingerprint` String CODEC(ZSTD(1)),
+    `seen_at_ts_bucket_start` Int64 CODEC(Delta(8), ZSTD(1))
+)
+ENGINE = Distributed('cluster', 'signoz_logs', 'logs_v2_resource_bucket', cityHash64(labels, fingerprint, seen_at_ts_bucket_start));
```
srikanthccv, line 20: Remove hardcoded cluster to `{{.SIGNOZ_CLUSTER}}`

srikanthccv, line 20: Why do we have `seen_at_ts_bucket_start` in sharding keys? If we just use a fingerprint it will guarantee the same fingerprint goes to the same shard.

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
```diff
@@ -0,0 +1,119 @@
...
+    `trace_id` String CODEC(ZSTD(1)),
+    `span_id` String CODEC(ZSTD(1)),
+    `trace_flags` UInt32,
+    `severity_text` LowCardinality(String) CODEC(ZSTD(1)),
+    `severity_number` UInt8,
+    `body` String CODEC(ZSTD(2)),
+    `attributes_string` Map(String, String) CODEC(ZSTD(1)),
+    `attributes_number` Map(String, Float64) CODEC(ZSTD(1)),
+    `attributes_bool` Map(String, Bool) CODEC(ZSTD(1)),
+    `resources_string` Map(String, String) CODEC(ZSTD(1)),
```
srikanthccv, line 39: Did we not want to use LowCarinality String for keys?

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
```diff
@@ -0,0 +1,119 @@
...
+    INDEX severity_text_idx severity_text TYPE set(25) GRANULARITY 4,
+    INDEX trace_flags_idx trace_flags TYPE bloom_filter GRANULARITY 4,
+    INDEX scope_name_idx scope_name TYPE tokenbf_v1(10240, 3, 0) GRANULARITY 4,
+    INDEX attributes_string_idx_key mapKeys(attributes_string) TYPE tokenbf_v1(1024, 2, 0) GRANULARITY 1,
+    INDEX attributes_string_idx_val mapValues(attributes_string) TYPE ngrambf_v1(4, 5000, 2, 0) GRANULARITY 1,
+    INDEX attributes_int64_idx_key mapKeys(attributes_number) TYPE tokenbf_v1(1024, 2, 0) GRANULARITY 1,
+    INDEX attributes_int64_idx_val mapValues(attributes_number) TYPE bloom_filter GRANULARITY 1,
+    INDEX attributes_bool_idx_key mapKeys(attributes_bool) TYPE tokenbf_v1(1024, 2, 0) GRANULARITY 1
+)
+ENGINE = MergeTree
```
srikanthccv, line 55: Same comment as above

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
```diff
@@ -0,0 +1,119 @@
...
+    `body` String CODEC(ZSTD(2)),
+    `attributes_string` Map(String, String) CODEC(ZSTD(1)),
+    `attributes_number` Map(String, Float64) CODEC(ZSTD(1)),
+    `attributes_bool` Map(String, Bool) CODEC(ZSTD(1)),
+    `resources_string` Map(String, String) CODEC(ZSTD(1)),
+    `scope_name` String CODEC(ZSTD(1)),
+    `scope_version` String CODEC(ZSTD(1)),
+    `scope_string` Map(String, String) CODEC(ZSTD(1))
+)
+ENGINE = Distributed('cluster', 'signoz_logs', 'logs_v2', cityHash64(id));
```
srikanthccv, line 83: Same

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
```diff
@@ -0,0 +1,119 @@
...
+
+
+ALTER TABLE signoz_logs.tag_attributes ON CLUSTER {{.SIGNOZ_CLUSTER}} modify column tagDataType Enum('string', 'bool', 'int64', 'float64', 'number') CODEC(ZSTD(1));
+ALTER TABLE signoz_logs.distributed_tag_attributes ON CLUSTER {{.SIGNOZ_CLUSTER}} modify column tagDataType Enum('string', 'bool', 'int64', 'float64', 'number') CODEC(ZSTD(1));
```
srikanthccv, line 87: Why do we have a `number`? It is not used. I see everything is float64. This is a mutation and should be avoided for now unless there is no strong reason

srikanthccv:
Can we export to v2 parallelly? also, need to update the instrumentation to record the time for each table.

Screenshot 2024-08-14 at 8 35 47 PM

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
ankitnayan: we should

srikanthccv:
Two more things I wanted to point out.

1. I noticed the table engine on customers instances is not replicated for just the v2 table (AFAIK at least two customers have it deployed).

2. I was playing around and looked at the queries. All of them use the `GLOBAL IN` for resource fingerprints and other Limit queries. I want to throw around the idea that if you can devise a fingerprinting and sharding mechanism that ensures the data is distributed evenly and the same fingerprint goes to the same shard, you can get rid of the `GLOBAL IN` which is not optimal compared to local IN. In its current shape I don't think data distribution will be even since it's entirely based on resource attributes and one set of resources can send disproportionately high data compared to others so we probably don't have an alternative but I want to bring it up anyway if you have ideas.

https://clickhouse.com/docs/en/sql-reference/operators/in

>When using GLOBAL IN / GLOBAL JOIN, first all the subqueries are run for GLOBAL IN / GLOBAL JOIN, and the results are collected in temporary tables. Then the temporary tables are sent to each remote server, where the queries are run using this temporary data.

>This will work correctly and optimally if you are prepared for this case and have spread data across the cluster servers such that the data for a single UserID resides entirely on a single server. In this case, all the necessary data will be available locally on each server. Otherwise, the result will be inaccurate. We refer to this variation of the query as “local IN”.

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
nityanandagohain: because there will be a comparison filter for `seen_at_ts_bucket_start` in all the queries.

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
nityanandagohain: Have added it for when we were about to shift but since we decided to move away for now will remove it.

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
srikanthccv: Not clear, how does comparison filter in query necessitate that it needs to be part of the sharding key?

nityanandagohain: Oh, my bad I took the sharding key for `order by`. 

Yeah right `seen_at_ts_bucket_start` doesn't seem to add any value to the sharding key. @raj-k-singh  any specific reason for adding this, else we can remove.

nityanandagohain:
Major changes made

- [x]  `logs_v2_resource_bucked` changed to `logs_v2_resource`
- [x] `span_attributes` is changed from array to string 
- [x] Write to both logs, logs_v2, logs_v2_resource table in  parallel
- [x] Separate duration metrics for all the above three tables

exporter/clickhouselogsexporter/logsv2/fingerprint.go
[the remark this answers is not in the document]
nityanandagohain: If we put as a synonym for `cloud.platform` then it will be a child of `gcp.project` . What we saw was that `source_type` was already present for all and some of them also had `gcp.project` . 

Do you see any issue with that?

migrationmanager/migrators/logs/migrations/000014_new_schema.up.sql
srikanthccv: How many examples have we seen? We should consider the general pattern and use lowcardinality. The special case users can be updated to use regular string if needed but not the other way around.

nityanandagohain: Not many users, yeah it will be better to tell them to reduce the attributes or if required then manually update for them.

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -321,18 +393,56 @@ func (e *clickhouseLogsExporter) pushToClickhouse(ctx context.Context, ld plog.L
...
+		for _, table := range []string{DISTRIBUTED_LOGS_TABLE, DISTRIBUTED_LOGS_TABLE_V2, DISTRIBUTED_LOGS_RESOURCE_V2} {
+			duration := <-chDuration
+			stats.RecordWithTags(ctx,
+				[]tag.Mutator{
+					tag.Upsert(exporterKey, component.DataTypeLogs.String()),
+					tag.Upsert(tableKey, table),
+				},
+				writeLatencyMillis.M(int64(duration.Milliseconds())),
+			)
+		}
```
srikanthccv, line 437: The order of the durations written to `chDuration` won't necessarily be the same as the order of the tables defined in iteration.

srikanthccv (approved):
Non-blocking point

srikanthccv (approved):

PR #353, opened 2024-07-23
fix: use maps for attribute and resources

nityanandagohain:
fix: use maps for attribute and resources


nityanandagohain:
closing this, addressed it here https://github.com/SigNoz/signoz-otel-collector/pull/348

PR #355, opened 2024-07-26
Skip write to time_series_v4 table for the same fingerprint in curren…

srikanthccv:
…t hour

Resubmission of https://github.com/SigNoz/signoz-otel-collector/pull/335

The previous version of the change considered only fingerprint but not time. It fails when the entry is added at the current hour :15 minutes. This resubmission includes the unixmilli as part of cache key. This will ensure there will be one fingerprint inserted if the hour changes.


nityanandagohain:
@srikanthccv  the code changes are the same, please add a description on why it was reverted and how we have fixed the issue because of which it's added back.

srikanthccv:
Updated!

nityanandagohain (approved):

PR #360, opened 2024-08-12
Move usage to new table write block

srikanthccv:
This change moves the counting samples for usage from `samples_v2` to `samples_v4`. There is no logical change. Just the code is moved to the new block.

### DIFF
--- a/exporter/clickhousemetricsexporter/clickhouse.go
+++ b/exporter/clickhousemetricsexporter/clickhouse.go
@@ -305,7 +305,6 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest, metr
 		return err
 	}
 
-	metrics := map[string]usage.Metric{}
 	err = func() error {
 		ctx := context.Background()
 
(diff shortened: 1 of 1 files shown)


exporter/clickhousemetricsexporter/clickhouse.go
```diff
@@ -359,12 +341,9 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest, metr
 		return err
 	}
 
-	for k, v := range metrics {
-		stats.RecordWithTags(ctx, []tag.Mutator{tag.Upsert(usage.TagTenantKey, k), tag.Upsert(usage.TagExporterIdKey, ch.exporterID.String())}, ExporterSigNozSentMetricPoints.M(int64(v.Count)), ExporterSigNozSentMetricPointsBytes.M(int64(v.Size)))
-	}
-
 	// write to distributed_samples_v4 table
 	if ch.writeTSToV4 {
```
nityanandagohain, line 345: this will always be true ?

nityanandagohain (approved):

exporter/clickhousemetricsexporter/clickhouse.go
srikanthccv: Yes https://github.com/SigNoz/signoz-otel-collector/blob/140c30d2c73e5925c6460107382d85c90b0edda6/exporter/clickhousemetricsexporter/factory.go#L128

PR #365, opened 2024-08-15
Chore/change resourcemap index

ankitnayan:
change the type of index for resource attributes from `bloom_filter` to `tokenbf_v1` for keys and `ngrambf_v1` for values. This is tested and implemented in logs already.

### DIFF
--- a/migrationmanager/migrators/traces/migrations/000031_change_resource_index.up.sql
+++ b/migrationmanager/migrators/traces/migrations/000031_change_resource_index.up.sql
@@ -1,0 +1,6 @@
+ALTER TABLE signoz_traces.signoz_index_v2 ON CLUSTER {{.SIGNOZ_CLUSTER}}
+    DROP INDEX IF EXISTS idx_resourceTagsMapKeys,
+    DROP INDEX IF EXISTS idx_resourceTagsMapValues,
+
+    ADD INDEX IF NOT EXISTS idx_resourceTagMapKeys mapKeys(resourceTagsMap) TYPE tokenbf_v1(1024, 2, 0) GRANULARITY 1,
+    ADD INDEX IF NOT EXISTS idx_resourceTagMapValues mapValues(resourceTagsMap) TYPE ngrambf_v1(4, 5000, 2, 0) GRANULARITY 1;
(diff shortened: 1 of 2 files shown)


ankitnayan:
Merge https://github.com/SigNoz/signoz-otel-collector/pull/367 before merging this

migrationmanager/migrators/traces/migrations/000031_change_resource_index.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE signoz_traces.signoz_index_v2 ON CLUSTER {{.SIGNOZ_CLUSTER}}
+    DROP INDEX IF EXISTS idx_resourceTagsMapKeys,
+    DROP INDEX IF EXISTS idx_resourceTagsMapValues,
```
srikanthccv, line 3: Can you drop the `DROP INDEX` from this migration and use different name for tokenbf indexes? `DROP INDEX` is implemented as mutation and will likely fail the migration and lead to dirty migration issues.

srikanthccv: We can DROP INDEXES later when we have the schema migrator ready.

ankitnayan: the names are different. I changed to keep the name synced to the column name. New is `idx_resourceTagsMapKeys` an old was `idx_resourceTagMapKeys`. Notice the `Tags` vs `Tag`.

Do you want to remove the `DROP INDEX ...` from up file only or also removing `Add INDEX ...` in the down file?

srikanthccv: Ah, I didn't notice that. Let's just remove the `DROP INDEX` from up migration file.

ankitnayan: 2 indexes on a column should be fine?

srikanthccv: Yes, it should be fine. We will incur some costs for index storage but from CH will pick the best one out of all created indexes.

srikanthccv (approved):
Will wait for other PR to get merged

ankitnayan:
> Merge #367 before merging this

this is not a blocker. The sequencing of migration files look fine

PR #367, opened 2024-08-15
chore: add index on traceID

ankitnayan:
adding index on traceID should enable searching by traceID much faster

### DIFF
--- a/migrationmanager/migrators/traces/migrations/000030_add_attributes_indices.down.sql
+++ b/migrationmanager/migrators/traces/migrations/000030_add_attributes_indices.down.sql
@@ -1,0 +1,8 @@
+ALTER TABLE signoz_traces.signoz_index_v2 ON CLUSTER {{.SIGNOZ_CLUSTER}}
+    DROP INDEX IF EXISTS idx_stringTagMapKeys,
+    DROP INDEX IF EXISTS idx_stringTagMapValues,
+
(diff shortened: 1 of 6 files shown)


ankitnayan:
this did not work in reducing granules scanned. Needs further investigation

nityanandagohain:
try using bloom filter instead of tokenbf. also increase the size of the filter

ankitnayan:
> try using bloom filter instead of tokenbf? also increase the size of the filter

Can you suggest the size to use?

srikanthccv:
@nityanandagohain please suggest the size. The https://github.com/SigNoz/signoz-otel-collector/pull/365 is waiting on this which will help in PRs like https://github.com/SigNoz/signoz/issues/3184

nityanandagohain:
If a granule has around 4000 unique trace_id ( 8192 total rows), 

The size of index = 10KB with 5 hash functions will give us probability of false positive 0.000314144 (1 in 3183).

https://hur.st/bloomfilter/?n=4000&p=&m=90000&k=5 

let's start with 10KB and 5 hash functions ?

ankitnayan:
this PR can wait

nityanandagohain:
https://github.com/SigNoz/signoz-otel-collector/blob/main/cmd/signozschemamigrator/schema_migrator/traces_migrations.go

PR #385, opened 2024-09-09
ci(build): set build workflow to use Primus workflow"

FavourDaniel:
The "push.yaml" workflow has been renamed "build.yaml". This workflow now uses the reusable Primus workflows.


grandwizard28:
Relates to https://github.com/SigNoz/platform-pod/issues/224

PR #389, opened 2024-09-11
Add column definition

srikanthccv:
This PR adds a `Column` that is used to describe the CH column in code. This will be used to implement all the operations on column.

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/column.go
+++ b/cmd/signozschemamigrator/schema_migrator/column.go
@@ -1,0 +1,305 @@
+package schemamigrator
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ColumnProperty represents a column property.
+// It is used to represent the column property in the column definition.
+// Example: TTL, DEFAULT, ALIAS, etc.
+type ColumnProperty string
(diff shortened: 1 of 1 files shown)


srikanthccv:
This should be easy one.

cmd/signozschemamigrator/schema_migrator/column.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	if c.Settings != nil {
+		sql.WriteString(" SETTINGS ")
+		sql.WriteString(c.Settings.String())
+	}
+	if c.Comment != "" {
+		sql.WriteString(" COMMENT ")
+		sql.WriteString(c.Comment)
+	}
+	return sql.String()
```
nityanandagohain, line 277: add few tests for these functions ?  or you want to address tests separately.

srikanthccv: The tests are coming in the upcoming PR where the column definition is used.

nityanandagohain (approved):
Looks good to start with,  we can add more if anything if required.

PR #392, opened 2024-09-11
Add TABLE operations

srikanthccv:
Support most of the operations mentioned in create table https://clickhouse.com/docs/en/sql-reference/statements/create/table

The `TableEngine` interface defines the funcs each table engine needs to implement. The following are supported table engines 

1. MergeTree
2. ReplacingMergeTree
3. AggregatingMergeTree
4. SummingMergeTree
and special distributed
5. Distributed

The operations on the table

1. CreateTableOperation
2. DropTableOperation
3. CreateMaterializedViewOperation

There are some operations we don't support not because they can't be supported but they aren't needed right now. They can be implemented where there is a need.

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/table.go
+++ b/cmd/signozschemamigrator/schema_migrator/table.go
@@ -1,0 +1,212 @@
+package schemamigrator
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TableEngine represents the engine of the table.
+type TableEngine interface {
+	ToSQL() string
+	EngineType() string
+	OnCluster(cluster string) TableEngine
+	WithReplication() TableEngine
+}
+
+// TableSetting represents the setting of the table.
+type TableSetting struct {
+	Name  string
+	Value string
+}
+
+func (t TableSetting) ToSQL() string {
+	return fmt.Sprintf("%s = %s", t.Name, t.Value)
+}
+
+type TableSettings []TableSetting
+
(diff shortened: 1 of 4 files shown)


cmd/signozschemamigrator/schema_migrator/table.go
```diff
@@ -0,0 +1,212 @@
...
+
+func (r ReplacingMergeTree) EngineType() string {
+	if r.Replicated {
+		return "ReplicatedReplacingMergeTree"
+	}
+	return "ReplacingMergeTree"
+}
+
+func (r ReplacingMergeTree) ToSQL() string {
+	fmt.Println("ReplacingMergeTree.ToSQL")
```
nityanandagohain, line 122: can be removed

cmd/signozschemamigrator/schema_migrator/table_test.go
```diff
@@ -0,0 +1,30 @@
...
+		wantSQL string
+	}{
+		{
+			name: "create table",
+			op: ReplacingMergeTree{
+				MergeTree{
+					OrderBy: "(timestamp, resource_id)",
+				},
+			},
+			wantSQL: "ReplacingMergeTree ORDER BY (timestamp, resource_id)",
```
nityanandagohain, line 20: nit: add one test with distributed, as for others like replicated it exists.

nityanandagohain (approved):
minor comments

PR #397, opened 2024-09-13
Add squashed migrations

srikanthccv:
The goal is to get rid of all `.sql` based operations and files. This changes creates a "squashed" migration containing the current snapshot of the schema

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/squashed_logs_migrations.go
+++ b/cmd/signozschemamigrator/schema_migrator/squashed_logs_migrations.go
@@ -1,0 +1,449 @@
+package schemamigrator
+
+var (
+	SquashedLogsMigrations = []SchemaMigrationRecord{
+		{
+			MigrationID: 1,
+			UpItems: []Operation{
+				CreateTableOperation{
+					Database: "signoz_logs",
+					Table:    "logs",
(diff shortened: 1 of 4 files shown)


cmd/signozschemamigrator/schema_migrator/squashed_logs_migrations.go
```diff
@@ -0,0 +1,685 @@
...
+					Table:    "distributed_usage",
+				},
+			},
+		},
+		{
+			MigrationID: 16,
+			UpItems: []Operation{
+				CreateTableOperation{
+					Database: "signoz_logs",
+					Table:    "logs_v2",
```
nityanandagohain, line 453: it's replicated by default for logs_v2 and logs_v2_resource

nityanandagohain (changes requested):

cmd/signozschemamigrator/schema_migrator/squashed_logs_migrations.go
srikanthccv: For every new user, replication will be enabled by default and controlled by that. It doesn't make any difference to hard code it here.

nityanandagohain (approved):

Issue #395, opened 2024-09-11
Add separate health endpoint for opamp managed collector


srikanthccv:
When a pipeline is updated it tries to restart the "service". The service restart first stops the receivers and attempts to flush the inmemory. The flush may take some time. However, since the collector has become unavailable the healthcheck extension returns service unavailable which leads to unwanted SIGKILL because the liveness probe fails.

PR #400, opened 2024-09-15
Add custom healthcheckextension

srikanthccv:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/395

All of the code is copied except the handler code in func `baseHandler` and `Handler()` is modified to return success all the time under the extension name `signoz_health_check`.


srikanthccv:
Even the current health check returns 200 all the time except when a shutdown is called. When shutdown is called, the extension becomes "NotReady" and returns 503 so that incoming data will be rejected and the client will retry. The same doesn't work for us because our tenant collectors are consuming from Kafka. If you are referring to the health of the pipeline, there is a v2 extension in development. In our case, a genuine health check failure is when we stop consuming from Kafka, and we have an alert for that. Does that answer the question?

srikanthccv:
Just to clarify, The problem we're facing is that while we wait for the pipeline data to be completely flushed, the extension is responding with a 503 error to the liveness probe, which triggers a kill eventually. We could extend the probe's timeout, but I thought the right solution would be to modify the consumer collector to return a 200 status code. This approach is preferable to arbitrarily increasing the probe time, as the latter might not be effective in all scenarios.

PR #402, opened 2024-09-16
Add sync/async sub command to schema migrator

srikanthccv:
1. Compared the existing schemas in staging and address the gaps
    - incorrect column name in logs
    - missing operations in metrics
    - incorrect ttl in traces
 2. when operation has replication enabled, the engine should be replaced with it's replicated version - addressed in this PR
 3. remove the localhost ref
 4. update the `waitForDistributionQueueOnHost` to fail if there are error items
 5. add migrations down code
 6. wait for mutations to complete before running the query as well.
 7. wire up all in the main with two sub commands `sync` and `async`.


srikanthccv:
@nityanandagohain please review

srikanthccv:
bump ^

nityanandagohain (approved):

PR #408, opened 2024-09-24
feat(kafka): apply backpressure

grandwizard28:
#### Features

- Apply backpressure on kafka on memorylimiter error. Pause the consumption and retry in place. This relives kafka of continuous rebalancing and gives a predictable memory workload for the collector

### DIFF
--- a/receiver/signozkafkareceiver/kafka_receiver.go
+++ b/receiver/signozkafkareceiver/kafka_receiver.go
@@ -149,8 +149,10 @@ func (c *kafkaTracesConsumer) Start(_ context.Context, host component.Host) erro
 		autocommitEnabled: c.autocommitEnabled,
 		messageMarking:    c.messageMarking,
 		baseConsumerGroupHandler: baseConsumerGroupHandler{
-			logger:        c.settings.Logger,
-			retryInterval: 1 * time.Second,
+			logger:          c.settings.Logger,
+			retryInterval:   1 * time.Second,
+			pausePartition:  make(chan int32),
+			resumePartition: make(chan int32),
 		},
 	}
 	go func() {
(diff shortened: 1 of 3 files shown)


srikanthccv:
Shouldn't we maintain one channel per partition instead of one channel for everything? Otherwise, other partitions would be blocking the channel without pausing which would trigger rebalance. Take the following case

1. collector got into high memory usage
2. then immediately clickhouse went down
3. the batch is being retried but the receiver would still get a high memory error
4. now only one partition has the pause but others are blocked
5. other partitions got stuck and do not report anything without pausing
6. kafka treats them as unhealthy and triggers rebalance.

grandwizard28:
I changed the pause to Pausing all partitions that the collector is currently subscribed to.
This should address the above issue.

receiver/signozkafkareceiver/kafka_receiver.go
```diff
@@ -400,7 +484,50 @@ func (c *kafkaLogsConsumer) Shutdown(context.Context) error {
...
+	if err == nil || !(err.Error() == errHighMemoryUsage.Error()) {
+		return err
+	}
+
+	// If errHighMemoryUsage is encountered, enter the retry loop
+	ticker := time.NewTicker(b.retryInterval)
+	defer ticker.Stop()
+
+	b.logger.Info("applying initial backpressure on Kafka due to high memory usage", zap.Int32("partition", claim.Partition()), zap.String("topic", claim.Topic()))
+	b.pausePartition <- claim.Partition()
```
srikanthccv, line 507: With the last commit of resuming and pausing all, the channel is no longer holding the partition but just a way to signal it. Please update it struct{}{} or something so that future readers won't get confused.

srikanthccv (commented):
LGTM. Minor comments.

receiver/signozkafkareceiver/kafka_receiver.go
grandwizard28: Resolved!

grandwizard28:
Added tests also and changed the errorLoop structure. It seemed too complex

srikanthccv (approved):

PR #411, opened 2024-09-28
Add new columns for separate point, scope, resource attributes

srikanthccv:


nityanandagohain (approved):
The changes are fine, but please double check if you want to go ahead with the name `attrs` instead of `attributes`.


in logs we use
`attributes`
`resources`

Issue #412, opened 2024-10-02
logs pipelines: new processor fails when deployed in collector pipelines with concurrent consumption


raj-k-singh:
The new processor runs into data races when used in collector pipelines with http server based receivers

PR #413, opened 2024-10-02
Fix: signozlogspipelineprocessor: support concurrency

raj-k-singh:
fixes https://github.com/SigNoz/signoz-otel-collector/issues/412


srikanthccv:
@raj-k-singh, can you help me understand where is the data race? It's unclear where the race is and how the change fixes it.

srikanthccv (approved):

PR #418, opened 2024-10-08
Bump upstream base version to 0.111.0

srikanthccv:


components/components.go
```diff
@@ -72,29 +95,52 @@ import (
...
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/sqlserverreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/sshcheckreceiver"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/statsdreceiver"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/syslogreceiver"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/tcplogreceiver"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/udplogreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/vcenterreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/wavefrontreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/webhookeventreceiver"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/windowseventlogreceiver"
```
nityanandagohain, line 142: is it necessary to add all of this in this receiver ?

srikanthccv: The unwritten policy we follow is: to include all the receiver/processor components so users can receive/process data from anywhere in the regular small self-hosted installations. But when it comes to exporters we only allow specific exporters.

nityanandagohain: ok

nityanandagohain (approved):

srikanthccv:
@raj-k-singh please take a look when you get some time.

PR #423, opened 2024-10-15
Trace details reference index changes

vikrantgupta25:


vikrantgupta25:
new PRs to be created when starting the implementation. this was the POC PR.

PR #430, opened 2024-10-18
clickhousetraces: Support specifying cluster

davispuh:
Currently `clickhousetraces` uses hardcoded cluster name `cluster` which doesn't allow to use different cluster.

This PR implements it as configuration option so any cluster can be specified.


davispuh:
By the way `clickhouseexporter` https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/clickhouseexporter#configuration-options supports `cluster_name` along with a lot of other configuration options.

srikanthccv (approved):

PR #436, opened 2024-10-24
feat: support for modify query mat col in migrations

### LINKED ISSUE #435: Running migration that has async and sync together.
* Let's say I have to create a migration, that has drop materialized view, and creating it again.
* Sync dropping it is a async operation and creating it is a sync operation what is the proper way to do it ?

nityanandagohain:
Fixes #435


srikanthccv (approved):

srikanthccv:
LGTM. Will double-check if it's implemented as a mutation.

PR #437, opened 2024-10-26
fix: panic due to concurrent map access & nil cache item in metric ex…

srikanthccv:
…porter

### DIFF
--- a/exporter/clickhousemetricsexporter/clickhouse.go
+++ b/exporter/clickhousemetricsexporter/clickhouse.go
@@ -424,8 +424,10 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest, metr
 
 			for fingerprint, labels := range timeSeries {
 				key := fmt.Sprintf("%d:%d", fingerprint, unixMilli)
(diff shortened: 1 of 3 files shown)


exporter/clickhousemetricsexporter/clickhouse.go
```diff
@@ -424,8 +424,10 @@ func (ch *clickHouse) Write(ctx context.Context, data *prompb.WriteRequest, metr
 
 			for fingerprint, labels := range timeSeries {
 				key := fmt.Sprintf("%d:%d", fingerprint, unixMilli)
-				if ch.cache.Get(key) != nil && ch.cache.Get(key).Value() {
-					continue
+				if item := ch.cache.Get(key); item != nil {
+					if value := item.Value(); value {
+						continue
+					}
```
srikanthccv, line 430: `if ch.cache.Get(key) != nil && ch.cache.Get(key).Value() { `

rarely the following happens: the check `ch.cache.Get(key) != nil` evals true, but then the cache entry expires and returns nil, then when `ch.cache.Get(key).Value()` is evaluated causing the nil pointer to dereference.

PR #447, opened 2024-11-06
feat: Trace v3 schema migration

nityanandagohain:
The schema migration for trace v3.

I have put it in the variable TracesMigrations_Staging as I don't want to run it right now.

Part of https://github.com/SigNoz/signoz/issues/5713

PR #449, opened 2024-11-08
feat: exporter for trace index_v3

nityanandagohain:
Follow up of https://github.com/SigNoz/signoz-otel-collector/pull/447 (please go through this first)

Part of https://github.com/SigNoz/signoz/issues/5713

* Here we are writing to signoz_index_v3 , trace_v3_resource, error span table and the tag attributes table.
* The billing calculation is done using the JSON representation of spanV3.
* It is not integrated i.e it won't run after getting merged
* We are not storing the static fields like `statusCode` in tag attributes as the query service will take care of it ( more details in later pr's of query service.

The reason I am calling this V3 is because of the table name signoz_index_v3

### DIFF
--- a/exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
+++ b/exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
@@ -100,6 +100,64 @@ func populateEventsV3(events ptrace.SpanEventSlice, span *SpanV3, lowCardinalExc
 	}
 }
 
+type attributesData struct {
+	StringMap      map[string]string
+	NumberMap      map[string]float64
+	BoolMap        map[string]bool
+	SpanAttributes []SpanAttribute
+}
+
+func (attrMap *attributesData) add(key string, value pcommon.Value) {
+	spanAttribute := SpanAttribute{
+		Key:      key,
+		TagType:  "tag",
+		IsColumn: false,
+	}
+
+	if value.Type() == pcommon.ValueTypeDouble {
+		attrMap.NumberMap[key] = value.Double()
+		spanAttribute.NumberValue = value.Double()
+		spanAttribute.DataType = "float64"
+	} else if value.Type() == pcommon.ValueTypeInt {
+		attrMap.NumberMap[key] = float64(value.Int())
+		spanAttribute.NumberValue = float64(value.Int())
+		spanAttribute.DataType = "float64"
+	} else if value.Type() == pcommon.ValueTypeBool {
+		attrMap.BoolMap[key] = value.Bool()
+		spanAttribute.DataType = "bool"
+	} else if value.Type() == pcommon.ValueTypeMap {
+		// flatten map
+		result := flatten.FlattenJSON(value.Map().AsRaw(), "")
+		for tempKey, tempVal := range result {
+			tSpanAttribute := SpanAttribute{
+				Key:      tempKey,
+				TagType:  "tag",
+				IsColumn: false,
+			}
+			switch tempVal := tempVal.(type) {
+			case string:
+				attrMap.StringMap[tempKey] = tempVal
+				tSpanAttribute.StringValue = tempVal
+				tSpanAttribute.DataType = "string"
+			case float64:
+				attrMap.NumberMap[tempKey] = tempVal
+				tSpanAttribute.NumberValue = tempVal
+				tSpanAttribute.DataType = "float64"
+			case bool:
+				attrMap.BoolMap[tempKey] = tempVal
+				tSpanAttribute.DataType = "bool"
+			}
+			attrMap.SpanAttributes = append(attrMap.SpanAttributes, tSpanAttribute)
+		}
+		return
+	} else {
+		attrMap.StringMap[key] = value.AsString()
+		spanAttribute.StringValue = value.AsString()
+		spanAttribute.DataType = "string"
+	}
+	attrMap.SpanAttributes = append(attrMap.SpanAttributes, spanAttribute)
+}
+
 func newStructuredSpanV3(bucketStart uint64, fingerprint string, otelSpan ptrace.Span, ServiceName string, resource pcommon.Resource, config storageConfig) (*SpanV3, error) {
 	durationNano := uint64(otelSpan.EndTimestamp() - otelSpan.StartTimestamp())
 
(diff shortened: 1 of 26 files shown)


exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -0,0 +1,324 @@
...
+			spanAttribute.NumberValue = float64(v.Int())
+			spanAttribute.DataType = "float64"
+		} else if v.Type() == pcommon.ValueTypeBool {
+			boolTagMap[k] = v.Bool()
+			spanAttribute.DataType = "bool"
+		} else {
+			stringTagMap[k] = v.AsString()
+			spanAttribute.StringValue = v.AsString()
+			spanAttribute.DataType = "string"
+		}
```
srikanthccv, line 146: My one issue with this is that it makes the maps unusable. For example, until recently, the clickhouse system table receiver sent a map in the attributes called `ProfilesEvents` with map values with keys and values. This is completely unusable from the UI. There is no point in storing if they can't be used from query builder. I believe we should flatten them here.
Screenshot 2024-11-10 at 6 50 23 PM

srikanthccv: Because the logs exporter was making them unusable strings. I had to flatten at the source https://github.com/SigNoz/signoz-otel-collector/pull/442. Ideally, the logs exporter itself should have flattened them for me.

nityanandagohain: Something like this is good https://go.dev/play/p/XVvXYASYFnu ?

srikanthccv: Yes, this looks good to me. Hopefully, the ClickHouse JSON type will soon be stable enough for use in such cases. Meanwhile, this is 100x better than the current unusable string.

nityanandagohain: Sure let me add it.

nityanandagohain: I have added the support for it, but not able to test it right now. Seems like the SDK's don't support maps as of now

```
Invalid type dict for attribute 'mydata' value. Expected one of ['bool', 'str', 'bytes', 'int', 'float'] or a sequence of those types
```

srikanthccv (commented):
Some comments. Please add tests.

exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
nityanandagohain: Yep, added.

exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -0,0 +1,346 @@
...
+		event.TimeUnixNano = uint64(events.At(i).Timestamp())
+		event.AttributeMap = map[string]string{}
+		event.IsError = false
+		events.At(i).Attributes().Range(func(k string, v pcommon.Value) bool {
+			event.AttributeMap[k] = v.AsString()
+			return true
+		})
+		if event.Name == "exception" {
+			event.IsError = true
+			span.ErrorEvent = event
```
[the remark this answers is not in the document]
nityanandagohain: added support for multiple exception events

exporter/clickhousetracesexporter/writerV3.go
```diff
@@ -0,0 +1,431 @@
...
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		err := w.writeTagBatchV3(ctx, batch)
+		if err != nil {
+			w.logger.Error("Could not write a batch of spans to tag/tagKey tables: ", zap.Error(err))
+			chErr <- err
+		}
```
srikanthccv, line 302: I wouldn't propagate the error from here since this is a metadata table. Failing the whole export because of the metadata table is not a good idea.

exporter/clickhousetracesexporter/writerV3.go
```diff
@@ -0,0 +1,431 @@
...
+				time.Unix(0, int64(errorEvent.Event.TimeUnixNano)),
+				errorEvent.ErrorID,
+				errorEvent.ErrorGroupID,
+				span.TraceId,
+				span.SpanId,
+				span.ServiceName,
+				errorEvent.Event.AttributeMap["exception.type"],
+				errorEvent.Event.AttributeMap["exception.message"],
+				errorEvent.Event.AttributeMap["exception.stacktrace"],
+				stringToBool(errorEvent.Event.AttributeMap["exception.escaped"]),
```
srikanthccv, line 116: could use [semconv](go.opentelemetry.io/collector/semconv) package.

exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -0,0 +1,360 @@
...
+	attrMap := attributesData{
+		StringMap:      make(map[string]string),
+		NumberMap:      make(map[string]float64),
+		BoolMap:        make(map[string]bool),
+		SpanAttributes: []SpanAttribute{},
+	}
+
+	resourceAttrs := map[string]string{}
+
+	otelSpan.Attributes().Range(func(k string, v pcommon.Value) bool {
```
srikanthccv, line 183: I do not know how many have nested resource attributes, but it would be nice to have flattened maps everywhere.

srikanthccv (approved):
Non-blocking comments.

nityanandagohain:
While making the changes for flattening resource attributes I also realized that they need not be stored in spanAttributes with a different dataType so updated the code. Previously we store resources in attributeMaps that's why it was added from what it seems.

exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -187,25 +187,29 @@ func newStructuredSpanV3(bucketStart uint64, fingerprint string, otelSpan ptrace
...
-		} else if v.Type() == pcommon.ValueTypeBool {
-			spanAttribute.DataType = "bool"
+		if v.Type() == pcommon.ValueTypeMap {
+			result := flatten.FlattenJSON(v.Map().AsRaw(), k)
+			for tempKey, tempVal := range result {
+				strVal := fmt.Sprintf("%v", tempVal)
+				resourceAttrs[tempKey] = strVal
+				attrMap.SpanAttributes = append(attrMap.SpanAttributes, SpanAttribute{
+					Key:         tempKey,
+					TagType:     "tag",
```
srikanthccv, line 197: Should be "resource"?

srikanthccv (approved):

PR #452, opened 2024-11-12
Move from OpenCensus to OTel API

srikanthccv:


nityanandagohain:
If not urgen't I will delay this review a bit.

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -96,17 +104,28 @@ func newExporter(logger *zap.Logger, cfg *Config) (*clickhouseLogsExporter, erro
 		return nil, err
 	}
 
+	durationHistogram, err := meter.Float64Histogram(
+		"exporter_db_write_latency",
+		metric.WithDescription("Time taken to write data to ClickHouse"),
+		metric.WithUnit("ms"),
+		metric.WithExplicitBucketBoundaries(250, 500, 750, 1000, 2000, 2500, 3000, 4000, 5000, 6000, 8000, 10000, 15000, 25000, 30000),
```
nityanandagohain, line 111: will 30s  max bucket be enough ? do we set a timeout for ingetion at 30s ?

nityanandagohain (approved):
overall looks good to me apart from the changes done in kafka receiver. Someone with context on that receiver might be better person to review. But unblocking from my side.

exporter/clickhouselogsexporter/exporter.go
srikanthccv: We have some cases with timeout set to 30s and default with 10s or something like that.

Will the 30s be enough? It's enough approximation to get an idea if exports are taking more than the sensible value. We already have a query log with exact timing so there is a better place to look for absolute time and how many rows are inserted in such export calls.

grandwizard28 (dismissed):

grandwizard28 (approved):

Issue #472, opened 2024-12-03
(logs/traces): do not write the same fingerprint to resource fingeprints table for same bucket ts


srikanthccv:
We know the ClickHouse is aggressive about ReplacingMergeTree merges; in exporter, for each batch, we write to resource table for the resource of the batch with bucket normalized time. This creates a new part and initiates a merge. If we see the same set of resources for a given bucket ts, there is no need to write to the resource table (should be written only when the buckets change or a new fingerprint is prepared). 

The implementation will maintain a cache of resource fingerprints and won't write any new rows if had been written already. The cache structure will set max capacity and ttl to protect against the unbounded growth.

srikanthccv:
Done!

PR #475, opened 2024-12-05
Add write cache for logs_v2_resource

srikanthccv:
logs part of https://github.com/SigNoz/signoz-otel-collector/issues/472

### DIFF
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -69,6 +69,7 @@ type clickhouseLogsExporter struct {
 	useNewSchema bool
 
 	keysCache *ttlcache.Cache[string, struct{}]
+	rfCache   *ttlcache.Cache[string, struct{}]
 }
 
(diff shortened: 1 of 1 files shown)


exporter/clickhouselogsexporter/exporter.go
```diff
@@ -445,11 +466,18 @@ func (e *clickhouseLogsExporter) pushToClickhouse(ctx context.Context, ld plog.L
 
 		for bucketTs, resources := range resourcesSeen {
 			for resourceLabels, fingerprint := range resources {
+				// if a resource fingerprint is seen in the bucket already, skip inserting it again.
+				key := makeKeyForRFCache(bucketTs, fingerprint)
+				if e.rfCache.Get(key) != nil {
```
nityanandagohain, line 471: Q: Will this update the ttl for the cache key when there is a hit ? if yes then it might be a problem as it won't be removed ever if the key is continiously

srikanthccv: See the init part
```
ttlcache.WithDisableTouchOnHit[string, struct{}](),
```

nityanandagohain (approved):

Issue #467, opened 2024-11-30
Do not write the tag_attributes/span_attributes tables if the number or unique values is above 25k


srikanthccv:
Free-form text attributes where every value is unique offer little practical value for storage or suggestions. When each entry is different from all others, maintaining these attributes or trying to generate recommendations becomes meaningless.

ankitnayan:
I agree to this. What would be the logic of implementation? Will the decision be made at the collector after doing some analysis on CH for attributes? It would be great if we can keep a few samples of the values for better UX but not necessarily

srikanthccv:
> What would be the logic of implementation? Will the decision be made at the collector after doing some analysis on CH for attributes?

Yes, the implementation would run a go routine periodically to fetch the distinct count of keys ingested from `signoz_logs.distributed_tag_attributes` and `signoz_traces.distributed_span_attributes`, next time all the keys that exceed the threshold will be skipped. To clarify my proposal is to apply these checks attributes of span/log, but not for the resource attribute gets ingested because they are usually controlled at the source. The additional condition that I believe we should apply is regardless of the distinct count, if a string value exceeds 256 (configurable) characters, they should also be skipped because there is no real-world scenario I can think of where people maintain such a lenghy enums/values that would make sense in suggestions.

> It would be great if we can keep a few samples of the values for better UX but not necessarily

I am not sure if it's going to be better UX if we consider all steps, imagine it suggested some values, but when you try to filter the search by typing something, no results show up because not all values are retained, it can confuse the user if the values exist or not. IMO, we should not show any suggestions for such keys (similar to what we do for the body). It should be either we fully retain all values and suggest (current status) or don't suggest anything (this issue).

ankitnayan:
okay... all of these makes sense

PR #477, opened 2024-12-06
Add logs tag attributes v2 with max distinct values limit

srikanthccv:
Logs part of https://github.com/SigNoz/signoz-otel-collector/issues/467

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/logs_migrations.go
+++ b/cmd/signozschemamigrator/schema_migrator/logs_migrations.go
@@ -80,4 +80,55 @@ ORDER BY name ASC`,
 			},
 		},
 	},
+	{
+		MigrationID: 1001,
+		UpItems: []Operation{
+			CreateTableOperation{
+				Database: "signoz_logs",
+				Table:    "tag_attributes_v2",
+				Columns: []Column{
+					{Name: "unix_milli", Type: ColumnTypeInt64, Codec: "Delta(8), ZSTD(1)"},
+					{Name: "tag_key", Type: ColumnTypeString, Codec: "ZSTD(1)"},
+					{Name: "tag_type", Type: LowCardinalityColumnType{ColumnTypeString}, Codec: "ZSTD(1)"},
+					{Name: "tag_data_type", Type: LowCardinalityColumnType{ColumnTypeString}, Codec: "ZSTD(1)"},
+					{Name: "string_value", Type: ColumnTypeString, Codec: "ZSTD(1)"},
+					{Name: "number_value", Type: NullableColumnType{ColumnTypeFloat64}, Codec: "ZSTD(1)"},
+				},
+				Indexes: []Index{
+					{Name: "string_value_index", Expression: "string_value", Type: "ngrambf_v1(4, 1024, 3, 0)", Granularity: 1},
+					{Name: "number_value_index", Expression: "number_value", Type: "minmax", Granularity: 1},
+				},
+				Engine: ReplacingMergeTree{
+					MergeTree: MergeTree{
+						PartitionBy: "toDate(unix_milli / 1000)",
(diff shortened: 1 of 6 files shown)


exporter/clickhouselogsexporter/exporter.go
```diff
@@ -614,13 +668,21 @@ func (e *clickhouseLogsExporter) addAttrsToAttributeKeysStatement(
...
 	statement driver.Batch,
+	tagStatementV2 driver.Batch,
 	attributeKeysStmt driver.Batch,
 	resourceKeysStmt driver.Batch,
 	tagType utils.TagType,
 	attrs attributesToSliceResponse,
 	useNewSchema bool,
+	shouldSkipKeys map[string]shouldSkipKey,
 ) error {
+	unixMilli := (time.Now().UnixMilli() / 3600000) * 3600000
```
nityanandagohain, line 679: any reason for unixmilli ? also what are you doing here ?

cmd/signozschemamigrator/schema_migrator/logs_migrations.go
```diff
@@ -80,4 +80,55 @@ ORDER BY name ASC`,
...
+				},
+				Indexes: []Index{
+					{Name: "string_value_index", Expression: "string_value", Type: "ngrambf_v1(4, 1024, 3, 0)", Granularity: 1},
+					{Name: "number_value_index", Expression: "number_value", Type: "minmax", Granularity: 1},
+				},
+				Engine: ReplacingMergeTree{
+					MergeTree: MergeTree{
+						PartitionBy: "toDate(unix_milli / 1000)",
+						OrderBy:     "(tag_key, tag_type, tag_data_type, string_value, number_value)",
+						TTL:         "toDateTime(unix_milli / 1000) + toIntervalSecond(1296000)",
```
nityanandagohain, line 105: Also why 15 days ? will it be same as table TTL ?

exporter/clickhouselogsexporter/exporter.go
srikanthccv: I need an epoch timestamp, rounded to the nearest 1 hour, and I generally use epoch milli. today, we store one instance for the key=value. The timestamp column contains the latest timestamp. In this new table, we maintain one entry for 1 hour. This will help, that otherwise can't be easily done with the existing table.

- What is the hourly churn rate?
- Show only relevant values in the selected period. 

Since we eliminate keys with high cardinality values from this table, storage won't be a problem.

cmd/signozschemamigrator/schema_migrator/logs_migrations.go
srikanthccv: Because fast filters don't work if they don't exist for two days https://github.com/SigNoz/engineering-pod/issues/1756. It won't be the same as the main table TTL. I think 15 should be good for both those who send data constantly and those who send data occasionally.

exporter/clickhouselogsexporter/exporter.go
nityanandagohain: can you add it as comment in code ? it will be helpful if someone looks at it.

cmd/signozschemamigrator/schema_migrator/logs_migrations.go
nityanandagohain: Ok

exporter/clickhouselogsexporter/exporter.go
srikanthccv: I intend to move this logic to the metadata exporter (https://github.com/SigNoz/signoz-otel-collector/issues/480, Do you have any concerns? I believe only main tables v2 and v2 resource logic should be part of the logs exporter), I will add it there.

nityanandagohain: ok

nityanandagohain (approved):

PR #479, opened 2024-12-07
Add metadata exporter

srikanthccv:

PR #482, opened 2024-12-11
Add unique value tracker

srikanthccv:
Part 2 of breakdown https://github.com/SigNoz/signoz-otel-collector/pull/479


srikanthccv:
Yes, the keys cache is ttl cache with max capacity and the values cache is lru with a fixed size.

PR #488, opened 2024-12-16
Add metadataexporter to components and reduce allocations

srikanthccv:


exporter/metadataexporter/exporter.go
```diff
@@ -69,7 +69,10 @@ type metadataExporter struct {
 func flattenJSONToStringMap(data map[string]any) map[string]string {
 	res := make(map[string]string, len(data))
 	for k, v := range data {
-		res[k] = fmt.Sprint(v)
+		switch v := v.(type) {
+		case string:
```
srikanthccv, line 73: We only support string so avoiding costly `fmt.Sprint`

exporter/metadataexporter/exporter.go
```diff
@@ -389,43 +395,47 @@ func (e *metadataExporter) shouldSkipAttributeUVT(_ context.Context, key, dataso
...
+		if typ == utils.TagDataTypeNumber.String() {
 			return true
 		}
 	}
 	return false
 }
 
+// filterAttrs filters attributes based on the unique value tracker
 func (e *metadataExporter) filterAttrs(ctx context.Context, attrs map[string]any, datasource string) map[string]any {
-	filtered := make(map[string]any, len(attrs))
```
srikanthccv, line 403: Avoiding allocation here by removing from the original map instead

nityanandagohain (approved):

PR #494, opened 2024-12-28
Add missing string value length limits

srikanthccv:


grandwizard28 (dismissed):

grandwizard28:
I was a little concerned about the attribute key limit. I looked at the messaging conventions and the largest key had 50 characters. We should be okay.

srikanthccv:
Removed the key length checks. If it ever becomes a problem, let it surface.

grandwizard28 (approved):

PR #493, opened 2024-12-28
feat(metering): add metering for logs

grandwizard28:
#### Features

- Add a pkg `metering` with interface `metering.Meter` for Size and Count usage calculation
- Add a pkg `pdatagen` for generating a wide variety of telemetry data for testing

PR #495, opened 2024-12-29
feat: Create a network payload meter

shivanshuraj1333:
Adds some refactoring on top of https://github.com/SigNoz/signoz-otel-collector/pull/493 which makes it generic and also allows it to be used in the gateway code.

Provides a way for serialization, it could be `json` or `proto`
Provides a way to do decoding, it could be `otlp` (plain upstream marshaling), or `signoz` (deep marshaling used in billing)

Implements `Meter` for `logs`, `metrics`, `traces`.

On the new instances of `logs`, `metrics`, `traces`, `Size` and `Count` can directly be called without worrying about anything.

But still there are few concerns, since we are deciding to put the module in `[signoz-otel-collector](https://github.com/SigNoz/signoz-otel-collector)` I am facing below problems:

- I am trying to import the new module (tagged with v0.111.19-68c19db8) but it is causing the `go.mod` tidy to get updated, we are using old libraries from collector, which is breaking the gateway code.

- This would cause the gateway code to break every time there's a library update in upstream.

- Even if we use this new package only without updating the other packges, it is importing other packages as a requirement mentioned in `go.mod` the reason gateway is importing new libraries is because of this https://github.com/SigNoz/signoz-otel-collector/pull/495/files#diff-b16c4c76ae88fb01ad12b932f6fa9fa83111a12fe0a996d208618fdfc75eb9f4R5-R10


grandwizard28:
Closing this as per discussion

PR #496, opened 2024-12-29
feat(metering): add meter for traces

grandwizard28:
#### Features

- The metering for traces is dependent on SigNoz's custom schema and not just on `ptrace.Traces`. The schema specific stuff has been moved to a separate package. It is not recommended to import the clickhousetracesexporter package in the metering package as it will bring a lot of unwanted dependencies into the codebase of whoever is importing it.
- Revert the modules created for metering and pdatagen. We thought we would import unwanted dependencies in the gateway which is not the case. We are only importing what we need there.

#### Fixes

- Remove redundant loggers in the LogsSize function.


grandwizard28:
@nityanandagohain @srikanthccv just want to double check my approach here. I'll push tests after that.

srikanthccv:
Is the size matching with the structuredspanv3?

srikanthccv:
I would suggest move structure span v3 to this package and use that instead.

grandwizard28:
Okay, 2 things:

- From a restructuring standpoint, I think it makes sense to move schema specific structs and their associated functions to a separate package. My recommendation is to move them to `pkg/schema/<signal>`.

- From the structuredSpanV3 standpoint, my initial thought was to move the entire NewStructuredV3Span function to the new package as mentioned in the point above. However, it seemed that a lot of functions would need to be moved. While I was fine with moving all function and was in the process of moving them, I realized a lot of the fields in SpanV3 have `json:"-"`. I did not want to waste computation on calculating them. The overhead of moving those functions + the wasted computation made me come up with this approach.

srikanthccv:
> I realized a lot of the fields in SpanV3 have json:"-". I did not want to waste computation on calculating them.

Which computation are you referring to? 

> The overhead of moving those functions

I think moving those functions is better than all the new piece-by-piece size functions introduced.

grandwizard28:
The custom columns we compute and the SpanAttributes for which we dont charge.

srikanthccv (approved):

PR #502, opened 2025-01-12
Add external cache for metadata exporter

srikanthccv:


exporter/metadataexporter/cache.go
```diff
@@ -0,0 +1,283 @@
...
+	var key strings.Builder
+	key.WriteString(tenantID)
+	key.WriteString(":metadata:")
+	key.WriteString(ds.String())
+	key.WriteString(":")
+	key.WriteString(strconv.FormatInt(currentEpochWindowMillis, 10))
+	return key.String()
+}
+
+type KeyCache interface {
```
grandwizard28, line 28: We should use a Pointer Receiver to implement this `KeyCache` to avoid copying the value on each method call. This can be more efficient if the receiver is a large struct (in the case of the InMemoryKeyCache implementation).

exporter/metadataexporter/cache.go
```diff
@@ -0,0 +1,283 @@
...
+	return InMemoryKeyCache{
+		tracesFingerprintCache:  tracesFingerprintCache,
+		metricsFingerprintCache: metricsFingerprintCache,
+		logsFingerprintCache:    logsFingerprintCache,
+		tenantID:                opts.TenantID,
+		logger:                  opts.Logger,
+	}
+}
+
+func getKeyForInMemory(tenantID string, ds pipeline.Signal, key string) string {
```
grandwizard28, line 88: This seems redundant now?

We can just use `getKey` for both.

grandwizard28 (changes requested):

exporter/metadataexporter/cache.go
grandwizard28:
```
newKey.WriteString(getKey(tenantID, ds))
```

Looking at this function, it seems like we are calling `getKey` as well. So the key seems to be:
`${getKey(tenantID, ds)}:${fingerprint}`

exporter/metadataexporter/cache.go
```diff
@@ -0,0 +1,283 @@
...
+	Password                   string
+	DB                         int
+	TracesFingerprintCacheTTL  time.Duration
+	MetricsFingerprintCacheTTL time.Duration
+	LogsFingerprintCacheTTL    time.Duration
+	TenantID                   string
+	Logger                     *zap.Logger
+}
+
+func NewRedisKeyCache(opts RedisKeyCacheOptions) *RedisKeyCache {
```
grandwizard28, line 167:
```suggestion
func NewRedisKeyCache(opts RedisKeyCacheOptions) (*RedisKeyCache, error) {
....
	if err := redisKeyCache.Ping(context.Background()).Err(); err != nil {
		return nil, err
	}
....
```

When we call `redis.NewClient`, it only initialises the struct and doesn't check whether the input details correspond to a valid redis connection or not. We should add a `Ping` to confirm.

exporter/metadataexporter/exporter.go
```diff
@@ -701,33 +728,22 @@ func (e *metadataExporter) PushLogs(ctx context.Context, ld plog.Logs) error {
...
 	}
 
-	e.set.Logger.Info("pushed logs attributes",
-		zap.Int("total_log_records", totalLogRecords),
-		zap.Int("skipped_log_records", skippedLogRecords),
-	)
-
-	if err := stmt.Send(); err != nil {
+	written, err := e.writeToStatementBatch(ctx, stmt, records, pipeline.SignalLogs)
+	if err != nil {
```
grandwizard28, line 743: I noticed we're not returning the error here by default. This is great for cases where redis failed.

Should we be more explicit and add a `fault_tolerant` mode to let users decide what should happen?

grandwizard28 (changes requested):

exporter/metadataexporter/exporter.go
srikanthccv: Throughout the exporter, it was a deliberate decision not to return errors. At this point, I don't see a good reason to return errors or even give an option.

exporter/metadataexporter/cache.go
srikanthccv: Yes, tenantID and ds are not needed but epochWindown is used.

exporter/metadataexporter/cache.go
srikanthccv: no error return policy in this exporter, but since this is the init stage, I will add it.

exporter/metadataexporter/exporter.go
grandwizard28: Okay!!!

grandwizard28 (approved):

PR #508, opened 2025-01-16
ci: support workflows for noninternal PRs

prashant-shahi:
We only want to run the primus CI workflows for the following cases:
- The PR is not from a fork and not from dependabot (due to security reasons, dependabot PRs are treated similar to that from Fork)
- The PR is labelled with "safe to test"

References:
- https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull-request-events-for-forked-repositories
- https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/


.github/workflows/remove-label.yaml
```diff
@@ -0,0 +1,16 @@
+name: remove-label
```
grandwizard28:
```suggestion
name: postci
```

grandwizard28 (approved):

grandwizard28 (approved):

PR #524, opened 2025-01-20
Chore: signozawsfirehosereceiver cwlog ts unmarshaling

raj-k-singh:
Cloudwatch log records contain timestamps in milliseconds.
OTLP log model expects timestamp to be in nanoseconds.

The forked code (from awsfirehose) was setting plog timestamp from the millisecond value leading to incorrect timestamps in parsed log records.

This change fixes that by correctly parsing the timestamp

contributes to https://github.com/SigNoz/signoz/issues/6540


srikanthccv:
Is it always going to be in millis? Would it be a good idea to handle any precision?

srikanthccv (approved):

PR #532, opened 2025-01-27
feat(signozclickhousemetrics): move to pointers

grandwizard28:
#### Description

- Use pointer to slices instead of slices.
- Rename to `signozclickhousemetrics`

#### Benchmark
```
type batchValue struct {
	samples []sample
	ts      []ts
}

type batchPointer struct {
	samples []*sample
	ts      []*ts
}

func generateSample() sample {
	return sample{
		env:         "production",
		temporality: pmetric.AggregationTemporalityCumulative,
		metricName:  "http_requests_total",
		fingerprint: 12345,
		unixMilli:   1632150000000,
		value:       123.45,
	}
}

func generateTs() ts {
	return ts{
		env:           "production",
		temporality:   pmetric.AggregationTemporalityCumulative,
		metricName:    "test",
		description:   "test",
		unit:          "test",
		isMonotonic:   false,
		attrs:         map[string]string{"a": "b"},
		scopeAttrs:    map[string]string{"a": "b"},
		resourceAttrs: map[string]string{"a": "b"},
	}
}

// BenchmarkGrowValue-1million    	       5	 232525983 ns/op	1149932720 B/op	      79 allocs/op
// BenchmarkGrowValue-10k 	     645	   1685683 ns/op	 8516823 B/op	      38 allocs/op
func BenchmarkGrowValue(b *testing.B) {
	s := generateSample()
	ts := generateTs()

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		batch := batchValue{
			samples: make([]sample, 0),
		}
		for j := 0; j < 10000; j++ {
			batch.samples = append(batch.samples, s)
			batch.ts = append(batch.ts, ts)
		}
	}
}

// BenchmarkGrowPointer-1million    	      14	  75,508,500 ns/op	89897213 B/op (85.732663)	      76 allocs/op
// BenchmarkGrowPointer-10k   	   15375	     77279 ns/op	  620785 B/op	      36 allocs/op
func BenchmarkGrowPointer(b *testing.B) {
	s := generateSample()
	ts := generateTs()

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		batch := batchPointer{
			samples: make([]*sample, 0),
		}
		for j := 0; j < 10000; j++ {
			batch.samples = append(batch.samples, &s)
			batch.ts = append(batch.ts, &ts)
		}
	}
}
```

With the struct we have, the `BenchmarkGrowPointer` is better suited.


exporter/clickhousemetricsexporterv2/exporter.go
```diff
@@ -751,43 +711,50 @@ func (c *clickhouseMetricsExporter) processExponentialHistogram(batch *writeBatc
...
 
-func (c *clickhouseMetricsExporter) prepareBatch(md pmetric.Metrics) *writeBatch {
-	batch := &writeBatch{metaSeen: make(map[string]struct{})}
+func (c *clickhouseMetricsExporter) prepareBatch(ctx context.Context, md pmetric.Metrics) *batch {
+	batch := newBatch()
 	start := time.Now()
 	for i := 0; i < md.ResourceMetrics().Len(); i++ {
 		rm := md.ResourceMetrics().At(i)
-		resAttrs := pcommon.NewMap()
-		rm.Resource().Attributes().CopyTo(resAttrs)
```
grandwizard28, line 769: All `CopyTo` calls have been removed. Very expensive.

exporter/clickhousemetricsexporterv2/internal/fingerprint.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+type Fingerprint struct {
+	attributes Attributes
+	hash       uint64
+	typ        FingerprintType
+}
+
+func NewFingerprint(typ FingerprintType, offset uint64, attrs pcommon.Map, extras map[string]string) *Fingerprint {
+	attributes := NewAttributesFromPcommonMap(attrs)
```
grandwizard28, line 22:
```suggestion
	attributes := NewAttributesFromPcommonMap(attrs)
```

This is happening the list number of times now.

exporter/clickhousemetricsexporterv2/exporter.go
srikanthccv: This doesn't make sense. How is `CopyTo` very expensive than creating a new map from `rm.Resource().Attributes()`?

exporter/clickhousemetricsexporterv2/exporter.go
```diff
@@ -291,52 +251,51 @@ func (c *clickhouseMetricsExporter) processGauge(batch *writeBatch, metric pmetr
...
 			typ:           typ,
 			isMonotonic:   isMonotonic,
-			fingerprint:   Fingerprint(pointAttrs, scopeAttrs, resAttrs, name),
+			fingerprint:   fingerprint.HashWithName(name),
 			unixMilli:     unixMilli,
-			labels:        getJSONString(getAllLabels(pointAttrs, scopeAttrs, resAttrs, name)),
-			attrs:         getAttrMap(pointAttrs),
-			scopeAttrs:    getAttrMap(scopeAttrs),
-			resourceAttrs: getAttrMap(resAttrs),
+			labels:        internal.NewLabels(name, fingerprint.AttributesAsMap(), scopeFingerprintMap, resourceFingerprintMap),
```
srikanthccv, line 275: I would keep the name as `getJSONString`. The `NewLabels` returns a string only be used for writing and is not usable beyond the part of `ts`. The `getJSONString` is more clear what it is doing.

exporter/clickhousemetricsexporterv2/exporter.go
```diff
@@ -348,54 +307,51 @@ func (c *clickhouseMetricsExporter) processSum(batch *writeBatch, metric pmetric
...
+			attrs:         fingerprint.AttributesAsMap(),
```
srikanthccv, line 331: Let's move the `fingerprint.AttributesAsMap()` out of struct and use it for both labels and attrs as it creates the same map both times.

srikanthccv:
@aniketio-ctrl please review

exporter/clickhousemetricsexporterv2/exporter.go
grandwizard28: `CopyTo` does unnecessary copying and rellocations under the hood if the input sizes don't match. Since the otlp protocol stores these as slices, `CopyTo` attempts to reallocate and create a new slice altogether which has significant overhead. See the `CopyTo` code:
```
	dest.getState().AssertMutable()
	newLen := len(*m.getOrig())
	oldCap := cap(*dest.getOrig())
	if newLen <= oldCap {
		// New slice fits in existing slice, no need to reallocate.
		*dest.getOrig() = (*dest.getOrig())[:newLen:oldCap]
		for i := range *m.getOrig() {
			akv := &(*m.getOrig())[i]
			destAkv := &(*dest.getOrig())[i]
			destAkv.Key = akv.Key
			newValue(&akv.Value, m.getState()).CopyTo(newValue(&destAkv.Value, dest.getState()))
		}
		return
	}

	// New slice is bigger than exist slice. Allocate new space.
	origs := make([]otlpcommon.KeyValue, len(*m.getOrig()))
	for i := range *m.getOrig() {
		akv := &(*m.getOrig())[i]
		origs[i].Key = akv.Key
		newValue(&akv.Value, m.getState()).CopyTo(newValue(&origs[i].Value, dest.getState()))
	}
	*dest.getOrig() = origs
```

exporter/clickhousemetricsexporterv2/exporter.go
grandwizard28: Ahh missed this

grandwizard28:
Also @srikanthccv, I would highly recommend (go as far as argue, make a strong case for) to change this exporter name to `signozclickhousemetrics`.

srikanthccv:
fine with the renaming, please update it.

exporter/clickhousemetricsexporterv2/exporter.go
srikanthccv: It creates a new slice of the len m and then copies the key value. I would imagine this to be as costly as us creating a map and assigning key value `NewAttributesFromPcommonMap`. Did you have a benchmark for this as well? Still not really sure how it is expensive in comparison.

grandwizard28: There's a benchmark with a negligible difference for this operation. (Since we are copying to an entirely new one so a new slice is created).

However, in general CopyTo bloats up especially when we want to add attributes to a map already containing attributes. That's why I mentioned very expensive. It was a general statement and a general thing I would like to avoid if we can get around it

srikanthccv (approved):

srikanthccv (approved):

Issue #534, opened 2025-01-29
log exporter mutates the data when the timestamp expands but doesn't set the mutates data capability


srikanthccv:
When we receive log records that create too many parts errors, we do in-place removal of data. However, the exporter doesn't set the capability to mutate data. This can lead to panic. We should either set the capability (which would mean it gets a new copy of data at the cost of additional memory) or refactor the logic to discard them differently.

If it were anywhere else in the pipeline, it would make sense to set the mutate capability, but since this is an exporter, the last point of the pipeline, we are anyway preparing the new log record for CH out of log record, I suggest refactoring it to discard from the list of already prepared logs record.

```
panic: invalid access to shared data

goroutine 2174 [running]:
go.opentelemetry.io/collector/pdata/internal.(*State).AssertMutable(...)
	/home/runner/go/pkg/mod/go.opentelemetry.io/collector/pdata@v1.17.0/internal/state.go:20
go.opentelemetry.io/collector/pdata/plog.LogRecordSlice.RemoveIf({0xc003624820?, 0xc00d7b2988?}, 0xf343b00?)
	/home/runner/go/pkg/mod/go.opentelemetry.io/collector/pdata@v1.17.0/plog/generated_logrecordslice.go:108 +0x126
github.com/SigNoz/signoz-otel-collector/exporter/clickhouselogsexporter.(*clickhouseLogsExporter).removeOldLogs(0xc004dcf180?, {0x9601ae0?, 0xc008344a10?}, {0xc007161d10?, 0xc00d7b2988?})
	/home/runner/work/signoz-otel-collector/signoz-otel-collector/exporter/clickhouselogsexporter/exporter.go:283 +0xea
github.com/SigNoz/signoz-otel-collector/exporter/clickhouselogsexporter.(*clickhouseLogsExporter).pushLogsData(0xc0027b46c0, {0x9601ae0, 0xc008344a10}, {0xc007161d10?, 0xc00d7b2988?})
	/home/runner/work/signoz-otel-collector/signoz-otel-collector/exporter/clickhouselogsexporter/exporter.go:304 +0x16f
go.opentelemetry.io/collector/exporter/exporterhelper.(*logsRequest).Export(0x9601a38?, {0x9601ae0?, 0xc008344a10?})
	/home/runner/go/pkg/mod/go.opentelemetry.io/collector/exporter@v0.111.0/exporterhelper/logs.go:61 +0x31
[...]

PR #536, opened 2025-01-30
fix: remove old logs while creating batch

nityanandagohain:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/534

We create a copy and then mutate the data and remove the old logs.


eKuG (dismissed):

srikanthccv:
@nityanandagohain This would have been fine if it was an occasional case, since we have several customers sending such data all the time, I think it would be best to remove them rather than letting the error happen and then retry by copying.

nityanandagohain:
@srikanthccv for that I will need the context of TTL, do you suggest getting the TTL every 10/20 mins or so  and store in in the cache ?

srikanthccv:
Yes, let's do that. Fetch on bootstrap and periodically check if there are any updates.

nityanandagohain:
@srikanthccv  let's say we removed logs which are older than the ttl, but still we get the error that too many partitions, what should we do ? drop the entire batch ?

srikanthccv:
Yes, drop and log you are dropping for this reason. Let it be the responsibility of the clickhouse service owner to set the `max_partitions_per_insert_block` as minimum as max TTL days. Since we are setting TTL manually, it will be on whoever is setting but in an automated process, the server setting would be updated automatically.

An alternative option is to make the collector get the setting max_partitions_per_insert_block and break down the batch into subbatches of max_partitions_per_insert_block days. That's too complex and shouldn't be the collector's job IMO.

nityanandagohain:
Have made the changes @srikanthccv , do have a look

srikanthccv:
@nityanandagohain LGTM, but ttl cache for just ttl seems like overkill, maybe just use atomic.Value?

nityanandagohain:
makes sense. I have updated it to use atomic.value

srikanthccv (approved):

PR #543, opened 2025-02-13
chore: signozeawsfirehosereceiver: cwmetrics: unmarshal to gauges instead of summary

raj-k-singh:
metrics stream JSON records like https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-metric-streams-formats-json.html get unmarshaled into summary right now
which is incorrect since summary is expected to have cumulative temporality and doesn't support delta while the metric stream sum and count values end up being deltas

This PR updates signozawsfirehosereceiver to ensure cwmetrics record types are unmarshaled into OTLP gauges - one gauge per statistic (count, sum, max, min)

contributes to https://github.com/SigNoz/signoz/issues/6540


srikanthccv (approved):
Created two issues for non-blocking changes https://github.com/SigNoz/signoz-otel-collector/issues/546 and https://github.com/SigNoz/signoz-otel-collector/issues/545. LGTM.

PR #548, opened 2025-02-18
fix(metrics): fix fingerprint for different order of same keys

srikanthccv:
This change

- Fixes the fingerprint generation to order key so that there will always be same fingerprint regardless of order of keys
- Add a __normalized column to indicate whether a row contains normalized metrics
- Add `__temporality__` to labels to use it to distinguish delta and cumulative


shivanshuraj1333:
can you please briefly describe what happens currently when keys are not sorted?

srikanthccv:
It creates a different fingerprint for the same set of labels, leading to unnecessary entries. This is not an issue in the current export being used. This is part of the new exporter code, and recent changes broke it.

harsh098:
Experienced the same issue with collector version v0.111.26, I have duplicate set of timeseries for same entries one with `__temporality__="Unspecified"` label and the other without them.

PR #554, opened 2025-03-05
feat(summary-update): added metric metadata table

aniketio-ctrl:
Added metric metadata table for storing updated metric type and metadata


cmd/signozschemamigrator/schema_migrator/squashed_metrics_migrations.go
```diff
@@ -1201,5 +1201,31 @@ FROM signoz_metrics.time_series_v4_1day;`,
...
+						OrderBy: "(metric_name)",
+					},
+				},
+			},
+			DownItems: []Operation{
+				DropTableOperation{
+					Database: "signoz_metrics",
+					Table:    "distributed_time_series_v4_1week",
+				},
+			},
```
srikanthccv, line 1228: Why are we dropping this?

cmd/signozschemamigrator/schema_migrator/squashed_metrics_migrations.go
```diff
@@ -1201,5 +1201,31 @@ FROM signoz_metrics.time_series_v4_1day;`,
 				},
 			},
 		},
+		{
+			MigrationID: 29,
+			UpItems: []Operation{
+				CreateTableOperation{
+					Database: "signoz_metrics",
+					Table:    "metrics_metadata",
```
srikanthccv, line 1209: We already have a table called "metadata" in the "signoz_metrics" DB. Should we name this updated_metadata or something similar?

cmd/signozschemamigrator/schema_migrator/squashed_metrics_migrations.go
```diff
@@ -1201,5 +1201,31 @@ FROM signoz_metrics.time_series_v4_1day;`,
...
+					Columns: []Column{
+						{Name: "metric_name", Type: ColumnTypeString, Codec: "ZSTD(1)"},
+						{Name: "temporality", Type: ColumnTypeString, Codec: "ZSTD(1)"},
+						{Name: "type", Type: ColumnTypeString, Codec: "ZSTD(1)"},
+						{Name: "description", Type: ColumnTypeString, Codec: "ZSTD(1)"},
+						{Name: "unit", Type: ColumnTypeString, Codec: "ZSTD(1)"},
```
srikanthccv, line 1215: All these can be LowCardinality

srikanthccv: We can keep it.

cmd/signozschemamigrator/schema_migrator/squashed_metrics_migrations.go
srikanthccv: I don't think `metrics_metadata` conveys its actual use. If you read the table name, would you interpret it as a table used for storing updated metadata?

aniketio-ctrl: nah, we can make it updated_metadata

cmd/signozschemamigrator/schema_migrator/squashed_metrics_migrations.go
srikanthccv: Please address this. And another thing missing in this PR is that every local table also has to have its corresponding distributed table. Let me know if it's not clear.

aniketio-ctrl: sharding key on metric_name?

srikanthccv: That should be fine or even rand() works. It might not matter much since the number of metrics is going to be small and doesn't matter which shard it goes to.

srikanthccv (approved):

srikanthccv (approved):

PR #561, opened 2025-03-17
fix(push): update if condition for without v-prefix build

prashant-shahi:
#### Fixes

- update if condition for without v-prefix build


vikrantgupta25 (approved):

grandwizard28:
Relates to https://github.com/SigNoz/platform-pod/issues/224

PR #570, opened 2025-03-22
fix(signozclickhousexeporter): added counter type for sum and count metrics in histogram

aniketio-ctrl:
We have added sum , counter , min and max metrics for histogram but their type is histogram only, it should be counter for sum, count and gauge for min, max.


srikanthccv:
You should update it for Summary metric type as well.

srikanthccv (approved):

PR #572, opened 2025-03-31
Fix/signoz-exporter: added extra attribute 

aniketio-ctrl:
Fix - 
the same labels and name was generated by the clickhousemetricsexporter so we added one more label to distinguish


srikanthccv:
@aniketio-ctrl, please explain how current instrumentation creates an issue and how adding another label with constant value solves it?

exporter/signozclickhousemetrics/exporter.go
```diff
@@ -813,6 +814,9 @@ func (c *clickhouseMetricsExporter) prepareBatch(ctx context.Context, md pmetric
 	c.processMetricsDuration.Record(
 		ctx,
 		float64(time.Since(start).Milliseconds()),
+		metricapi.WithAttributes(
+			attribute.String("exporter", exporterName),
+		),
```
srikanthccv, line 819: Let's not have mixed usage, i.e., `exporter` and `exporter.name,` which have different meanings for different metrics.

aniketio-ctrl:
> @aniketio-ctrl, please explain how current instrumentation creates an issue and how adding another label with constant value solves it?

this is because both the exporter is sending same metrics with same label name that what causing the issue, thats why i have added one more label to distinguish

exporter/signozclickhousemetrics/exporter.go
aniketio-ctrl: no i have added one more label exporter.name only for exporter_db_write_latency, only exporter label would be fine for processmetricduration

srikanthccv: Right, I was referring to inconsistency in the label use. The `exporter` in our custom metric refers to signal kind. But in this case, it is the exporter's name. Please create a separate issue for addressing the instrumentation labels consistency.

srikanthccv:
Thanks for the explanation. Please fix CI.

exporter/signozclickhousemetrics/exporter.go
aniketio-ctrl: ok ill create diff issue for that

grandwizard28:
Part of https://github.com/SigNoz/platform-pod/issues/621

PR #581, opened 2025-04-03
fix(metrics-exporter): added nan check in metrics exporter | 7472

aniketio-ctrl:
We are skipping the rows for which Nan values are there, as OpenTelemetry proactively drops NaN values at the SDK level to ensure that only valid data is processed and visualized, so we should not consider those rows at all.


exporter/signozclickhousemetrics/exporter.go
```diff
@@ -505,6 +513,11 @@ func (c *clickhouseMetricsExporter) processHistogram(b *batch, metric pmetric.Me
 		// 3. min
 		// 4. max
 		// 5. bucket counts
+		// Check if any of the key float fields are NaN.
+		if math.IsNaN(dp.Sum()) || math.IsNaN(dp.Min()) || math.IsNaN(dp.Max()) {
+			c.logger.Warn("NaN detected in data point, skipping entire data point", zap.String("metric_name", name))
+			continue
+		}
```
srikanthccv, line 520: Why do we not have checks for count and buckets?

exporter/signozclickhousemetrics/exporter.go
```diff
@@ -617,6 +630,10 @@ func (c *clickhouseMetricsExporter) processSummary(b *batch, metric pmetric.Metr
 		// 1. count
 		// 2. sum
 		// 3. quantiles
+		if math.IsNaN(dp.Sum()) {
+			c.logger.Warn("NaN detected in data point, skipping entire data point", zap.String("metric_name", name))
+			continue
+		}
```
srikanthccv, line 636: Same why no checks for count and quantile?

exporter/signozclickhousemetrics/exporter.go
```diff
@@ -767,6 +784,10 @@ func (c *clickhouseMetricsExporter) processExponentialHistogram(b *batch, metric
 		// 3. min
 		// 4. max
 		// 5. ddsketch
+		if math.IsNaN(dp.Sum()) || math.IsNaN(dp.Min()) || math.IsNaN(dp.Max()) {
+			c.logger.Warn("NaN detected in data point, skipping entire data point", zap.String("metric_name", name))
+			continue
+		}
```
srikanthccv, line 790: Same

exporter/clickhousemetricsexporter/helper.go
```diff
@@ -332,10 +334,14 @@ func addSingleNumberDataPoint(pt pmetric.NumberDataPoint, resource pcommon.Resou
 	case pmetric.NumberDataPointValueTypeDouble:
 		sample.Value = pt.DoubleValue()
 	}
+	if math.IsNaN(sample.Value) {
+		return ErrNaNDetected
+	}
 	if pt.Flags().NoRecordedValue() {
 		sample.Value = math.Float64frombits(value.StaleNaN)
 	}
```
srikanthccv, line 342: In this line, we are setting the value to be certain NaN so it's possible that this will mess with the aggregations.

exporter/signozclickhousemetrics/exporter.go
aniketio-ctrl: because they are uint64 and cant be nan

exporter/signozclickhousemetrics/exporter.go
aniketio-ctrl: because they are uint64 and cant be nan or inf

exporter/signozclickhousemetrics/exporter.go
aniketio-ctrl: because they are uint64 and cant be nan

exporter/clickhousemetricsexporter/helper.go
aniketio-ctrl: yeah we can drop this row if detect no recorded value

exporter/signozclickhousemetrics/exporter.go
srikanthccv: https://github.com/hashicorp/consul/issues/4254#issuecomment-457621583

exporter/clickhousemetricsexporter/helper.go
aniketio-ctrl: but we need to have this stale nan for prom ql queries otherwise it will show old values, not consider that series as dead

srikanthccv: So what do we do?

aniketio-ctrl: The challenge is that we need StaleNaN (or some marker) for PromQL queries to correctly interpret series as stale, but we can't store NaN values in ClickHouse due to aggregation constraints.

One solution is to track which series have the NoRecordedValue flag by marking is_stale : true, and when feeding data into the PromQL engine, we can manually inject a StaleNaN for those series at the appropriate timestamp.

srikanthccv:
@aniketio-ctrl, please add tests.

pkg/pdatagen/pmetricsgen/data.go
```diff
@@ -385,3 +388,377 @@ func GenerateSummaryMetrics(numMetrics, numDataPoints, numPointAttributes, numSc
 	}
 	return metrics
 }
+
+func GenerateGaugeNanMetrics(numMetrics, numDataPoints, numPointAttributes, numScopeAttributes, numResourceAttributes int) pmetric.Metrics {
```
srikanthccv, line 392: Refactor this to make `NaN` and `NonRecordedValue`, `Nil` as `GenerationOption`

srikanthccv (commented):
You should test this against the real DB instance at least once. Please share how was this tested.

pkg/pdatagen/pmetricsgen/data.go
```diff
@@ -24,105 +50,92 @@ func Generate(opts ...GenerationOption) pmetric.Metrics {
...
-			dp := gaugeMetric.DataPoints().AppendEmpty()
-			dp.SetTimestamp(pcommon.NewTimestampFromTime(time.Now()))
-			dp.SetDoubleValue(rand.Float64())
-		}
-	}
+	// Generate and copy metrics
+	copyMetricsToScope(GenerateGaugeMetrics(
+		generationOpts.count.GaugeMetricsCount,
+		generationOpts.count.GaugeDataPointCount,
+		0,
```
srikanthccv, line 71: Add `generationOptions.pointAttributeCount` to generationOptions and use it everywhere instead random zero, which is more readable and configurable. This random zero breaks code reading flow and one has to read GenerateGaugeMetrics signature to unerstand what the zero maps to.

pkg/pdatagen/pmetricsgen/data.go
```diff
@@ -181,47 +218,41 @@ func GenerateGaugeMetrics(numMetrics, numDataPoints, numPointAttributes, numScop
...
 		}
+
+		dpSlice := sum.DataPoints()
+		validPoints := numDataPoints - numNoRecordedValue
+
+		// Add normal data points
+		addDataPoints(dpSlice, i, validPoints, numPointAttributes, baseTimestamp, "sum.attr_", false)
+
+		// Add NoRecorded data points
+		addDataPoints(dpSlice, i, numNoRecordedValue, numPointAttributes, baseTimestamp, "sum.attr_", true)
```
srikanthccv, line 253: NaN for sum missing

pkg/pdatagen/pmetricsgen/data.go
```diff
@@ -133,40 +146,64 @@ func Generate(opts ...GenerationOption) pmetric.Metrics {
...
-		for j := 0; j < numDataPoints; j++ {
+		validDataPoints := numDataPoints - numNanValues - numNoRecordedValues
+
+		// Add normal data points
+		addDataPoints(dpSlice, i, validDataPoints, numPointAttributes, baseTimestamp, "gauge.attr_", false)
+
+		// Add NoRecorded data points
+		addDataPoints(dpSlice, i, numNoRecordedValues, numPointAttributes, baseTimestamp, "gauge.attr_", true)
+
+		// Add NaN data points
```
srikanthccv, line 170: Why not move this to `addDataPoints` like `addSummaryDataPoints`?

pkg/pdatagen/pmetricsgen/data.go
```diff
@@ -133,40 +146,64 @@ func Generate(opts ...GenerationOption) pmetric.Metrics {
...
+
 	for i := 0; i < numMetrics; i++ {
 		m := sm.Metrics().AppendEmpty()
 		m.SetName("system.memory.usage" + strconv.Itoa(i))
 		m.SetUnit("bytes")
 		m.SetDescription("memory usage of the host")
+
 		dpSlice := m.SetEmptyGauge().DataPoints()
-		for j := 0; j < numDataPoints; j++ {
+		validDataPoints := numDataPoints - numNanValues - numNoRecordedValues
```
srikanthccv, line 162: This would be surprising, without knowing the impl detail one would imagine they get back metrics with `numDataPoints` valid point, `numNanValues`, nan points, and `numNoRecordedValues` no rec points but this. Let's either rename the `numDataPoints` to `numValidDataPoints` or make the func return `numDataPoints + numNanValues + numNoRecordedValues` points

srikanthccv (commented):
Some final comments

PR #584, opened 2025-04-11
fix: JSON Parser to accept value as a map

piyushsingariya:


nityanandagohain:
Can you check if there is any way to add tests for the changes you added ?

processor/signozlogspipelineprocessor/stanza/operator/operators/json/parser_test.go
```diff
@@ -98,6 +98,36 @@ func TestParser(t *testing.T) {
...
+			&entry.Entry{
+				Attributes: map[string]any{
+					"superkey": "superval",
+					"superkey2": map[string]any{
+						"nestedkey1": "nestedval",
+					},
+					"superkey3": []string{"val1", "val2"},
+				},
+				// Note: body is not strigified by parser but the exporter when saving in clickhouse
+				Body: map[string]any{
```
nityanandagohain, line 122: add a case for map[string]string , so that we are sure.

piyushsingariya: Tested and every map based payload is received as map[string]any.

Even payload sent as map[int]float64 is received as map[string]any

nityanandagohain: then it's good enough

nityanandagohain (approved):

nityanandagohain (approved):

PR #593, opened 2025-05-02
add wget to container

leonowski:
In the documents for setting up ECS infra monitoring, a healthcheck option is used.

https://signoz.io/docs/userguide/collecting-ecs-sidecar-infra/

The healthcheck option uses wget.  Previous releases used alpine as the base image and alpine has wget through busybox.  This PR is to add wget to the container so that the ECS healthcheck works without error.


JohnNeptune:
This would be appreciated unless anyone has an alternative way of checking that the container itself is healthy?

PR #597, opened 2025-05-16
exprimental/test: update Upstream packages

piyushsingariya:
### User Story

A user of SigNoz has reported flaky behavior in the SigNoz Logs Pipeline where resource tags from some logs are incorrectly attached to other logs. The SigNoz OpenTelemetry Collector code processes each log independently, and we suspect the issue is not caused there.

In earlier versions, the pkg/stanza component operated asynchronously, which might have contributed to this issue. To investigate, we are raising this experimental PR to update the upstream OTEL packages and observe if the problem persists.

For more context, see the related Slack discussion: https://signoz-team.slack.com/archives/C02B85R4R0A/p1746709316367769


opamp/server_client.go
```diff
@@ -119,27 +119,39 @@ func (s *serverClient) Start(ctx context.Context) error {
 		return err
 	}
 
+	instanceID := func(s string) types.InstanceUid {
```
nityanandagohain, line 122: please add comment what you are doing here

nityanandagohain (commented):
a minor comment apart from that looks good to me , @srikanthccv  have a look once

PR #611, opened 2025-06-16
fix: ignore random keys in the tag attributes and attribute keys table

nityanandagohain:
Added check to see if a keys is worth storing in the tag attributes and attribute keys table.

Part of https://github.com/SigNoz/engineering-pod/issues/2128

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
+++ b/cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
@@ -804,18 +804,6 @@ var MetricsMigrations = []SchemaMigrationRecord{
 	{
 		MigrationID: 1004,
 		UpItems: []Operation{
-			DropTableOperation{
-				Database: "signoz_metrics",
-				Table:    "time_series_v4_6hrs_mv_separate_attrs",
-			},
-			DropTableOperation{
-				Database: "signoz_metrics",
-				Table:    "time_series_v4_1day_mv_separate_attrs",
-			},
-			DropTableOperation{
-				Database: "signoz_metrics",
-				Table:    "time_series_v4_1week_mv_separate_attrs",
-			},
 			ModifyQueryMaterializedViewOperation{
 				Database: "signoz_metrics",
 				ViewName: "time_series_v4_6hrs_mv",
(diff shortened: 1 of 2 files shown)


pkg/keycheck/keycheck_test.go
```diff
@@ -0,0 +1,149 @@
...
+		{
+			name: "String with Vowels",
+			key:  "aeiouAEIOU",
+			want: false,
+		},
+		{
+			name: "Mixed Case without Digits",
+			key:  "Abcdefghijklmnop",
+			want: true,
+		},
```
srikanthccv, line 50: This could be a problem because data from AWS could easily be like this (they use CapitalCase, and keys can be lengthy)

pkg/keycheck/keycheck.go
```diff
@@ -0,0 +1,112 @@
...
+			case hexRegex.MatchString(segment):
+				return true
+			case base64Regex.MatchString(segment):
+				return true
+			case timestampRegex.MatchString(segment):
+				return true
+			case len(segment) > 12 && !containsVowels(segment):
+				return true
+			case len(segment) > 16 && hasUpperLowerDigit(segment):
+				return true
```
srikanthccv, line 58: I think we should remove these two. It would be risky to reject the legitimate keys.

pkg/keycheck/keycheck_test.go
nityanandagohain: hmm, so 256 length limit should just be good right ?

srikanthccv: By lengthy, i mean long enough to trigger the following logic.

```
			case len(segment) > 12 && !containsVowels(segment):
				return true
			case len(segment) > 16 && hasUpperLowerDigit(segment):
				return true

```

nityanandagohain: is around 30 a good number for a segment then ?

srikanthccv: Yes, that sounds good to me.

nityanandagohain:
cc @piyushsingariya for your reference.

Issue #601, opened 2025-05-22
Issue deploying signoz-otel-collector in IPv6 environment


Nic0w:
Hi,

I am trying to deploy SigNoz in a k8s single-stack IPv6 cluster using Helm charts (charts.signoz.io).
After a few changes to accomodate for the IPv6-only environment, everything starts up except the signoz-otel-collector container.

First there is a CrashLoopBackoff state for the signoz-schema-migrator-sync-init container with the following logs :
```
{"L":"info","timestamp":"2025-05-22T17:11:50.943Z","C":"signozschemamigrator/main.go:91","M":"Running migrations in sync mode","dsn":"tcp://admin:27ff0399-0d3a-4bd8-919d-17c2181e6fb9@signoz-clickhouse:9000","replication":false,"cluster-name":"cluster"}
{"L":"info","timestamp":"2025-05-22T17:11:50.945Z","C":"signozschemamigrator/main.go:104","M":"Up migrations","versions":[]}
{"L":"info","timestamp":"2025-05-22T17:11:50.945Z","C":"signozschemamigrator/main.go:117","M":"Down migrations","versions":[]}
{"L":"info","timestamp":"2025-05-22T17:11:50.945Z","C":"signozschemamigrator/main.go:127","M":"Parsed DSN","optsError":"json: unsupported type: func(context.Context, string) (net.Conn, error)"}
{"L":"info","timestamp":"2025-05-22T17:11:50.946Z","C":"signozschemamigrator/main.go:133","M":"Opened connection"}
{"L":"info","timestamp":"2025-05-22T17:11:51.531Z","C":"schema_migrator/manager.go:152","M":"Created databases","dbs":["signoz_traces","signoz_metrics","signoz_logs","signoz_metadata","signoz_analytics"]}
{"L":"info","timestamp":"2025-05-22T17:11:51.533Z","C":"schema_migrator/manager.go:161","M":"Creating schema migrations tables"}
{"L":"info","timestamp":"2025-05-22T17:11:51.533Z","C":"schema_migrator/manager.go:873","M":"Running operation","migration_id":1,"database":"signoz_logs","skip_status_update":true}
{"L":"info","timestamp":"2025-05-22T17:11:51.534Z","C":"schema_migrator/manager.go:883","M":"Waiting for running mutations before running the operation"}
[...]

PR #616, opened 2025-06-18
fix: distinguish IPv4 from IPv6 addresses (#601)

Nic0w:
This is an attempt to provide a fix for issue #601.

The problem seems to lie in function HostAddrs at line 341 where host address is concatenated to port
https://github.com/SigNoz/signoz-otel-collector/blob/1d7bf04531ca0e64d490c630e2a6090ba5ee9e37/cmd/signozschemamigrator/schema_migrator/manager.go#L334-L341

Then later in the for loop at line 345 it tries to open connections (line 349):
https://github.com/SigNoz/signoz-otel-collector/blob/1d7bf04531ca0e64d490c630e2a6090ba5ee9e37/cmd/signozschemamigrator/schema_migrator/manager.go#L345-L350

I never wrote any Go code but I think the fix might be quite simple. According to [this StackOverflow answer](https://stackoverflow.com/questions/22751035/golang-distinguish-ipv4-ipv6), testing the result of  `To4()` would help distinguish IPv4 and IPv6 addresses. 

Please note this is my first time writing Go. I'd appreciate if someone more knowledgeable would pick this up =)
This fix would allow me to deploy Signoz in an IPv6-only environment.

Thanks again for any help

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/manager.go
+++ b/cmd/signozschemamigrator/schema_migrator/manager.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"sync"
 	"time"
+	"net"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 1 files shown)


grandwizard28:
Appreciate the PR @Nic0w, you were somewhat close to what we ended up doing:
https://github.com/SigNoz/signoz-otel-collector/pull/644

PR #626, opened 2025-06-23
feat: Start collector with NopReceiver

piyushsingariya:
fixes SigNoz/platform-pod#906 

Starts collector with Noop Configuration before connecting and receiving remote configuration from OpAMP Server

### DIFF
--- a/opamp/server_client.go
+++ b/opamp/server_client.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/SigNoz/signoz-otel-collector/constants"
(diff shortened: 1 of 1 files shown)


grandwizard28 (dismissed):

nityanandagohain (changes requested):
I tested it locally
* delete signoz db and start signoz and collector
* collector starts with noop
* create org and apply pipelines
* the new changes are saved but not applied to collector i.e collector still running with noop.
* If I restart the collector it does pick up the new change.

@piyushsingariya  in the last second step in the background it should try to keep connecting to the signoz server with a backoff and update the config as soon as the connection is successful without restarting

grandwizard28 (dismissed):

nityanandagohain (approved):

PR #649, opened 2025-07-15
chore: remove __resource.schema_url__ from metric labels

srikanthccv:


srikanthccv:
The OTEL specification doesn't state explicitly if the Resource schema URL is a metric identifying attribute. It has been noticed that the multiple values for `__resource.schema_url__` create different unique series even if the rest of the attributes are the same. This leads to incorrect results. To address this problem, we are removing the resource schema URL from the series identity. What about the Scope info then? This section of the compatibility says the scope name, version and URL must be added as labels to series when exposing with PromEx. This indicates they are series identifying attributes so we don't make any change for the scope attributes https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#instrumentation-scope-1

aniketio-ctrl (approved):

PR #660, opened 2025-08-07
feat(meter): exclude signoz internal resource attributes from usage

vikrantgupta25:


exporter/clickhousetracesexporter/schema-signoz.go
```diff
@@ -166,7 +166,8 @@ type SpanV3 struct {
 	AttributesNumber map[string]float64 `json:"attributes_number,omitempty"`
 	AttributesBool   map[string]bool    `json:"attributes_bool,omitempty"`
 
-	ResourcesString map[string]string `json:"resources_string,omitempty"`
+	ResourcesString         map[string]string `json:"_,omitempty"`
+	BillableResourcesString map[string]string `json:"resources_string,omitempty"`
```
vikrantgupta25, line 170: had to keep the key same so that no bill diff incurrs

internal/common/signoz_attributes.go
```diff
@@ -0,0 +1,7 @@
+package common
```
grandwizard28, line 1: Please move this to `pkg/metering`.

grandwizard28 (changes requested):

Issue #677, opened 2025-09-05
Keys appear to be written only once for the entire period of collector process


srikanthccv:
Some internal reports and OSS reports indicate that the keys are not getting written to the table periodically. They get written only once during the entire collector process lifetime. This means, if the collector is running without restart for 15 days, users start to see "Key not found" error. We should fix the issue.

https://signoz-team.slack.com/archives/C08N7DQUNMS/p1756986842643349
https://signoz-community.slack.com/archives/C01HWQ1R0BC/p1756722678672369

PR #679, opened 2025-09-08
fix: log attribute and resource keys not present

nityanandagohain:
Fixes https://github.com/SigNoz/signoz-otel-collector/issues/677


srikanthccv:
@nityanandagohain I see its default value is set to 240 minutes already https://github.com/SigNoz/signoz-otel-collector/blob/421ade220563bfee55cb33a0b5c4873e62ccce36/exporter/clickhouselogsexporter/factory.go#L76C3-L76C54. Can we check why it's not working?

srikanthccv:
keys cache also doesn't have `WithDisableTouchOnHit`, does that affect?

nityanandagohain:
No it doesn't matter in this case as here `addAttrsToAttributeKeysStatement` was not even called and was put under shouldSkipKeys.

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -639,13 +639,13 @@ func (e *clickhouseLogsExporter) addAttrsToTagStatement(
 			e.logger.Debug("attribute value length exceeds the limit", zap.String("key", attrKey))
 			continue
 		}
+		e.addAttrsToAttributeKeysStatement(attributeKeysStmt, resourceKeysStmt, attrKey, tagType, utils.TagDataTypeString)
```
srikanthccv, line 642: Let's move this above the value length check on 638 line?

nityanandagohain: But do we want to pollute our resource table as well ? there might be some wrong pipeline applied and because of which it will pollute the keys table as well.

srikanthccv: I didn't mean to allow the random keys. On 638, we are checking if the length of value >= 256 and continue. I can see a case where users have some keys with such long strings and wants to search on them. I think we should have such keys, even if the value is long. However, we don't want to have auto-complete on them which is reasonable.

PR #684, opened 2025-09-19
fix: ignore spans with empty span id in the MVs

nityanandagohain:
Ignore spans which doesn't have a span id, both on the left and right side of join.

Fixes https://github.com/SigNoz/engineering-pod/issues/2982


nityanandagohain:
@srikanthccv 

For  service calls mv
```
4 rows in set. Elapsed: 0.030 sec. Processed 89.63 thousand rows, 7.05 MB (2.96 million rows/s., 232.49 MB/s.)
Peak memory usage: 5.96 MiB.
```

Sub root operations
```
4 rows in set. Elapsed: 0.015 sec. Processed 89.63 thousand rows, 1.58 MB (5.97 million rows/s., 105.42 MB/s.)
Peak memory usage: 1.36 MiB.
```

Previously both of the query took around 11 minutes. This is without adding any max_threads or query_cache.

PR #689, opened 2025-10-04
feat(manager): add a retry loop for no node coordination error

grandwizard28:
#### Features

When clickhouse DDLWorker deletes the task before we query the distributed_ddl_queue, a error is thrown saying "No Node: Code: 999. Coordination::Exception: Coordination error: No node, path \/clickhouse\/signoz-clickhouse\/task_queue\/ddl\/query-0000001757."

We are adding a retry loop to prevent crashes of schema migrator when this happens


grandwizard28:
Part of https://github.com/SigNoz/platform-pod/issues/1216

PR #695, opened 2025-10-27
feat: add normalize operator for log  (PR 1: JSON Logs)

piyushsingariya:
This PR makes sure, that if there is a string escaped JSON present, it considers it as a JSON so that it can be used further down the way.

This PR intends to break down the OG pr #694 into smaller piece containing only Log Body preprocessor

### DIFF
--- a/processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/transformer.go
+++ b/processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/transformer.go
@@ -2,7 +2,6 @@ package json
 
 import (
 	"context"
-	"sort"
 	"strings"
 
 	signozstanzahelper "github.com/SigNoz/signoz-otel-collector/processor/signozlogspipelineprocessor/stanza/operator/helper"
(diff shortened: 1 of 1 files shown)


processor/signozlogspipelineprocessor/stanza/operator/operators/preprocessing/transformer.go
```diff
@@ -0,0 +1,79 @@
...
+	default:
+		return nil // return nil to avoid erroring out the pipeline
+	}
+
+	// set parsed value to body
+	entry.Body = parsedValue
+
+	messageField := signozstanzaentry.NewBodyField("message")
+	// add first found msg compatible field to body
+	for _, fieldName := range msgCompatibleFields {
```
nityanandagohain: Is this something we really want to do ? i.e mapping something without informing user, as if we think about informing then user will have to again get the context that we map `msg` to `message`.

nityanandagohain (commented):
Are these points the goal of the operator.

* Convert stringified json body to a map type.
* Map `msg` to `message`

Also since this will be just an operator processor for SigNoz json, should we rename it to signozjson operator ?

I have a little doubt about mapping `msg` to `message` automatically and should be a decision taken by the user though config in the UI (inside or outside pipeline). And should this be converted to a standard config where user can do more things in the preprocessing itself.

nityanandagohain (dismissed):

PR #696, opened 2025-10-27
feat: JSON type exporter (PR 2: JSON Logs)

piyushsingariya:
This PR is broken down component of main PR #694 only containing the JSON type exporter

Closes SigNoz/engineering-pod#2989

```
goos: darwin
goarch: arm64
pkg: github.com/SigNoz/signoz-otel-collector/exporter/jsontypeexporter
cpu: Apple M4 Pro
BenchmarkPushLogs_10k-12    	      49	  21710063 ns/op	12558155 B/op	  372713 allocs/op
PASS
ok  	github.com/SigNoz/signoz-otel-collector/exporter/jsontypeexporter	3.412s
```


exporter/jsontypeexporter/exporter.go
```diff
@@ -0,0 +1,333 @@
...
+				e.keyCache.Add(key, struct{}{}) // add key to cache to avoid checking it again for next log records
+			}
+
+			path := prefix + "." + key
+			if prefix == "" {
+				path = key
+			} else if inArray {
+				path = prefix + key
+			}
+			if err := e.analyzePValue(ctx, path, false, value, typeSet); err != nil {
```
nityanandagohain: a config should be there to limit how much depth this should visit.

exporter/jsontypeexporter/exporter.go
```diff
@@ -0,0 +1,333 @@
...
+		typeSet.Insert(prefix, maskInt)
+		return nil
+	default:
+		return fmt.Errorf("unknown type at path: %s", prefix)
+	}
+}
+
+// persistTypes writes the collected types to the ClickHouse database
+func (e *jsonTypeExporter) persistTypes(ctx context.Context, typeSet *TypeSet, flush bool) error {
+	if !flush && typeSet.counter.Load() < 10_000 {
```
nityanandagohain: this means if I send a few log lines and then stop sending it won't be written ?

shouldn't it be a independent go routine ?

nityanandagohain (commented):
Added some comments.

will `jsonlogmetadata`  be a better name?

I will look more into the concurrency code and the ploggen.

exporter/jsontypeexporter/exporter.go
piyushsingariya: removed this

piyushsingariya:
> will `jsonlogmetadata` be a better name?

I think one right now is explicitly adhering the task

pkg/keycheck/cardinality.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// containsDigits checks if string contains digits
+func containsDigits(s string) bool {
+	if len(s) == 0 {
+		return false
+	}
+
+	// Check if all characters are digits
+	for _, char := range s {
```
nityanandagohain: During testing phase check existing queries which have number in attributes

PR #705, opened 2025-11-08
chore: address linting issues

srikanthccv:


cmd/signozschemamigrator/schema_migrator/manager.go
```diff
@@ -368,7 +368,7 @@ func (m *MigrationManager) HostAddrs() ([]string, error) {
 	if err != nil {
 		return nil, errors.Join(ErrFailedToGetHostAddrs, err)
 	}
-	defer rows.Close()
+	defer func() { _ = rows.Close() }()
```
piyushsingariya: We can add .Close() pattern to be ignored in `.golangci.yml`??

Issue #698, opened 2025-10-28
V2 span metrics with connector


srikanthccv:
- Names of the new metrics
- Metric types of the new metric
- How does it play with upstream connector

nikhilmantri0902:
This was deprioritised.

PR #708, opened 2025-11-09
feat(signozspanmetricsconnector): addition of traces to metrics connector Part-2

nikhilmantri0902:
Part of #698


connectors/signozspanmetricsconnector/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+type: signozspanmetrics
+
+status:
```
nikhilmantri0902: can we remove this file? @srikanthccv

PR #717, opened 2025-11-17
chore(metadataexporter): change log level from info to debug for limits

therealpandey:
### Summary
change log level from info to debug for limits


srikanthccv (approved):
This would make us blind to who are actually hitting limits, but right now we are anyway not acting on who is hitting the limits so we are fine for now. I will create a separate ticket to either make it metric or some less noisy version but still show up warn logs.

PR #727, opened 2025-12-01
feat: adding ColumnTypeEnum (JSON PR 5)

piyushsingariya:
### Summary
This PR adds ColumnTypeEnum and minor changes in JSONColumnType; These will be used in JSON's PRs
This fixes an issue with compare types in switch case, earlier we directly used to compare with structs but due to upgrade in ColumnTypes now that is not possible hence, adding ColumTypeEnums

### Related Issues

Closes #

### Resource Impact
- [ ] Yes
- [x] No


piyushsingariya:
> Also since you are going to use enum, will something like iota make more sense when declaring the constants instead of assigning the same string again ?

make sense, changing into that

PR #730, opened 2025-12-03
feat: JSON Log Body Migration (JSON PR 6)

piyushsingariya:
### Summary
This PR includes migration changes required for JSON logs

### Related Issues

Closes #

### Resource Impact
- [x] Yes
- [ ] No

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/manager.go
+++ b/cmd/signozschemamigrator/schema_migrator/manager.go
@@ -854,6 +854,11 @@ func (m *MigrationManager) MigrateUpAsync(ctx context.Context, upVersions []uint
 			continue
 		}
 		for _, item := range migration.UpItems {
+			if item.ForceMigrate() {
+				m.logger.Info("Skipping force sync operation", zap.Uint64("migration_id", migration.MigrationID))
+				// skip the force sync operation (should run in sync mode)
+				continue
+			}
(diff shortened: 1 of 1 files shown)


cmd/signozschemamigrator/schema_migrator/logs_migrations_v2.go
```diff
@@ -0,0 +1,441 @@
...
+					Name: constants.BodyJSONColumn,
+				},
+			},
+			InsertIntoTable{
+				Database:    SignozMetadataDB,
+				Table:       constants.DistributedPromotedPathsTable,
+				LightWeight: true,
+				Synchronous: true,
+				Columns:     []string{"path", "created_at"},
+				Values:      fmt.Sprintf("('message', %d)", time.Now().UnixMilli()),
```
nityanandagohain: time.now() will not be idempotent i.e you won't known when it was added if this migration is run again which is `INSERT INTO %s.%s (path, created_at) VALUES ('message', %d)` will just insert the data with new timestamp and the latest time will be noted, better to make sure it is not replaced in the query?

nityanandagohain (changes requested):
* Missed checking ForceMigrate in upItems of MigrateUpAsync in metrics and logs
* Same in MigrateDownAsync

PR #734, opened 2025-12-17
chore: added migration and code for column inserted_at_unix_milli for samples_v4 and time_series_v4 tables

### LINKED ISSUE #729: Add inserted_at column for samples_v4 and time_series_v4 tables
I initially pushed against the idea to save some space (~50gb for 100 billion samples). But a lot of time spent time in analysing the false alerts could simply be explained if we know exactly when the sample is stored on the database vs when it is produced on client side. 

At that, scale additional 50gb should be fine for the time it will save us.

nikhilmantri0902:
Closes #729 
### Summary
Briefly describe the change and its purpose.

### Related Issues

Closes #

### Resource Impact
- [ ] Yes
- [ ] No

If **Yes**, describe:
  
### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
+++ b/cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
@@ -955,6 +955,24 @@ var MetricsMigrations = []SchemaMigrationRecord{
 					Codec: "ZSTD(1)",
 				},
 			},
+			AlterTableAddColumn{
+				Database: "signoz_metrics",
+				Table:    "exp_hist",
+				Column: Column{
+					Name:  "inserted_at_unix_milli",
+					Type:  ColumnTypeInt64,
+					Codec: "ZSTD(1)",
+				},
+			},
+			AlterTableAddColumn{
+				Database: "signoz_metrics",
+				Table:    "distributed_exp_hist",
+				Column: Column{
+					Name:  "inserted_at_unix_milli",
+					Type:  ColumnTypeInt64,
+					Codec: "ZSTD(1)",
+				},
+			},
 		},
 	},
 }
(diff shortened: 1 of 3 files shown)


cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
```diff
@@ -916,4 +916,46 @@ var MetricsMigrations = []SchemaMigrationRecord{
...
 	},
+	{
+		MigrationID: 1007,
+		UpItems: []Operation{
+			AlterTableAddColumn{
+				Database: "signoz_metrics",
+				Table:    "samples_v4",
+				Column: Column{
+					Name:  "inserted_at_unix_milli",
+					Type:  NullableColumnType{ColumnTypeInt64},
```
srikanthccv: Why is it a nullable column?

nikhilmantri0902: As we wont be able to rightly backfill ingestion timestamps for past values, IMO it should be kept as null for these values.

srikanthccv: That's not the right way to decide the type of the column. The backfill is a separate thing. The type of the column should be based on the nature of the data. When a column is nullable, it means that by nature there can be cases where there won't be a value with the underlying type and should be null to indicate the missing value. Does that apply to `inserted_at_unix_milli`? No, because there is never a case where `inserted_at_unix_milli` is unknown, so it doesn't make sense to make the type nullable. Now, coming to the backfill question, if you leave it unfilled, ClickHouse returns a zero value which can be used to know whether something is old data or new data.

Additionally, nullable columns in ClickHouse also have a small storage overhead (null bitmap) and can complicate queries (needing isNotNull() checks, potential for unexpected behavior with aggregations).

srikanthccv:
- please check the ci failure
- please add the same col for the exp hist table

nikhilmantri0902:
> * please check the ci failure
> * please add the same col for the exp hist table

This was done. Lint failing is not related to this change.

PR #736, opened 2025-12-21
Bump collector version and contrib components to v0.142.0

nikhilmantri0902:
Closes https://github.com/SigNoz/engineering-pod/issues/3480

### Summary
Briefly describe the change and its purpose.

`Note`: This PR will drop `carbonexporter`, `opencensusexporter`, `routingprocessor`, `opencensusreceiver`, because upstream 0.142.0 no longer ships these packages. Replacing the per-component modules with the 0.142.0 root failed because the packages are absent, and the older per-component tags (carbon v0.140.1, opencensus* v0.133.0, routing v0.132.0) don’t compile against the new collector APIs (configoptional, Host ToServer/ToClient changes, pointer OTTL contexts). 

The following are the changes needed in the collector config : 
1. Changes needed for otel-collector image to run
` --feature-gates=pkg.translator.prometheus.NormalizeName`
2. In collector config, remove enabled option from otel-collector-config
`sending_queue: {}`

### DIFF
--- a/cmd/signozotelcollector/config/config.go
+++ b/cmd/signozotelcollector/config/config.go
@@ -9,12 +9,14 @@ import (
 )
 
 var (
-	Collector        collector
-	Clickhouse       clickhouse
-	MigrateReady     migrateReady
-	MigrateBootstrap migrateBootstrap
-	MigrateSyncCheck migrateSyncCheck
-	MigrateSyncUp    migrateSyncUp
+	Collector         collector
+	Clickhouse        clickhouse
+	MigrateReady      migrateReady
+	MigrateBootstrap  migrateBootstrap
+	MigrateSyncCheck  migrateSyncCheck
+	MigrateSyncUp     migrateSyncUp
+	MigrateAsyncCheck migrateAsyncCheck
+	MigrateAsyncUp    migrateAsyncUp
 )
 
 type collector struct {
@@ -40,7 +42,7 @@ type clickhouse struct {
 func (cfg *clickhouse) RegisterFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&cfg.DSN, "clickhouse-dsn", "tcp://0.0.0.0:9001", "DSN for clickhouse connection")
 	cmd.PersistentFlags().StringVar(&cfg.Cluster, "clickhouse-cluster", "cluster", "Name of the clickhouse cluster to connect")
-	cmd.PersistentFlags().BoolVar(&cfg.Replication, "clickhouse-replication", false, "Set true if replication is enabled in the clickhouse cluster")
+	cmd.PersistentFlags().BoolVar(&cfg.Replication, "clickhouse-replication", true, "Set true if replication is enabled in the clickhouse cluster")
 }
 
 type migrateReady struct {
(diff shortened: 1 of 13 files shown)


exporter/clickhouselogsexporter/config_test.go
```diff
@@ -61,12 +62,17 @@ func TestLoadConfig(t *testing.T) {
...
+		QueueBatchConfig: configoptional.Some(exporterhelper.QueueBatchConfig{
 			Sizer:        exporterhelper.RequestSizerTypeRequests,
 			NumConsumers: 10,
 			QueueSize:    100,
-		},
+			Batch: configoptional.Default(exporterhelper.BatchConfig{
+				FlushTimeout: 200 * time.Millisecond,
+				Sizer:        exporterhelper.RequestSizerTypeItems,
+				MinSize:      8192,
+				MaxSize:      0,
```
srikanthccv: Why is MaxSize zero?

exporter/clickhouselogsexporter/metadata.yaml
```diff
@@ -4,6 +4,7 @@ status:
   stability:
     stable: [logs]
 
+# ttlcache starts a background worker; cleanup stops it, but we ignore it as a safety net
```
srikanthccv: it says cleanup stops it and yet ignores it. why?

exporter/signozclickhousemeter/config_test.go
```diff
@@ -15,6 +18,15 @@ func TestConfig_Validate(t *testing.T) {
...
 		DSN: "tcp://localhost:9000?database=default",
+		QueueBatchConfig: configoptional.Some(exporterhelper.QueueBatchConfig{
+			QueueSize:    100,
+			NumConsumers: 1,
+			Batch: configoptional.Default(exporterhelper.BatchConfig{
+				FlushTimeout: 200 * time.Millisecond,
+				Sizer:        exporterhelper.RequestSizerTypeItems,
+				MinSize:      8192,
+			}),
+		}),
```
srikanthccv: Why is this added here?

pkg/collectorsimulator/collectorsimulator.go
```diff
@@ -146,6 +148,8 @@ func NewCollectorSimulator(
 		ExtensionsConfigs:   collectorCfg.Extensions,
 		ExtensionsFactories: factories.Extensions,
 
+		TelemetryFactory: factories.Telemetry,
```
srikanthccv: What is this for?

srikanthccv:
@piyushsingariya can you please review the pipeline components related changes?

srikanthccv:
Sorry missed the last message.

> Are we discussing about data operations/validations failing and we jump to next pipeline (we just log them), that's our internal implementation of Stanza using an external lib to form expressions and run data validation through that (sort of like evaluation), if that passes we do the operation and if those fails we move on. lmk if I am in wrong direction

No, I am referring to the breaking change above in `signoztransformprocessor`, will that breaking change result in collector bootstrap failures if we roll out new version? i.e or in other words, do pipelines work with signontranform processor, if so, do we let users configure pipelines from UI in way that this breaking change will be problematic.

srikanthccv (approved):

PR #742, opened 2025-12-30
fix: JSON collector fixes

### LINKED ISSUE #739: Unsupported value types shouldn't return error to the upstream component
The JSON type exporter should not return errors to the previous pipeline component for unsupported types in body json, as these permanent failures trigger infinite, futile retries. Instead, errors should only be returned for transient failures that warrant retry attempts.

### LINKED ISSUE #740: Divergence between the json exporter types and telemetry types mapping
https://github.com/SigNoz/signoz/blob/f6da9adb86a752083d0c8a0becc1e9d3c2fd1164/pkg/types/telemetrytypes/json_datatype.go#L35-L47

https://github.com/SigNoz/signoz-otel-collector/blob/c51e67fd26be1ace6c0dc89db43779a4f97ce47e/exporter/jsontypeexporter/types.go#L9-L20

piyushsingariya:
### Summary
This PR contains fixes in collector for JSON logs testing

### Related Issues

Closes #740 
Closes #739 

### Resource Impact
- [ ] Yes
- [ ] No

If **Yes**, describe:
  
### DIFF
--- a/exporter/jsontypeexporter/exporter.go
+++ b/exporter/jsontypeexporter/exporter.go
@@ -249,7 +249,7 @@ func (e *jsonTypeExporter) analyzePValue(ctx context.Context, val pcommon.Value,
 					cur = maskArrayInt
 				default:
 					// move to next element
-					e.logger.Error("unknown element type in array at path", zap.String("path", prefix))
+					e.logger.Error("unknown element type in array at path", zap.String("path", prefix), zap.Any("type", el.Type()))
 				}
 				if i > 0 && cur != prev {
 					mixed = true


exporter/jsontypeexporter/exporter.go
```diff
@@ -139,6 +139,16 @@ logIteration:
 	return nil
 }
 
+func (e *jsonTypeExporter) pushLogs(ctx context.Context, ld plog.Logs) error {
+	if err := e.processLogs(ctx, ld); err != nil {
+		if e.config.FailOnError {
+			return err
+		}
+	}
```
srikanthccv: missed the log here?

exporter/jsontypeexporter/exporter.go
```diff
@@ -226,7 +236,8 @@ func (e *jsonTypeExporter) analyzePValue(ctx context.Context, val pcommon.Value,
 					}
 					cur = maskArrayJSON
 				case pcommon.ValueTypeSlice:
-					return fmt.Errorf("arrays inside arrays are not supported! found at path: %s", prefix)
+					// pass via the array element
+					e.logger.Error("arrays inside arrays are not supported!", zap.String("path", prefix))
```
srikanthccv: This can now return an error.

exporter/jsontypeexporter/exporter.go
piyushsingariya: oh shoot.

exporter/jsontypeexporter/exporter.go
srikanthccv: That makes sense. There are other errors that break the loop; let's address them as well?

https://github.com/SigNoz/signoz-otel-collector/blob/c51e67fd26be1ace6c0dc89db43779a4f97ce47e/exporter/jsontypeexporter/exporter.go#L239

srikanthccv (approved):

PR #760, opened 2026-01-31
test: add tests for checking migration sanity

therealpandey:
### Summary
- add tests for checking migration sanity
- a single migration needs to have all sync or all sync operations. These tests ensure we check this at compile time.

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/data_operations_test.go
+++ b/cmd/signozschemamigrator/schema_migrator/data_operations_test.go
@@ -21,11 +21,11 @@ func TestInsertIntoTable(t *testing.T) {
 			name: "insert-into-table",
 			op: InsertIntoTable{
 				Database: constants.SignozMetadataDB,
-				Table:    constants.DistributedPromotedPathsTable,
-				Columns:  []string{"path", "created_at"},
-				Values:   fmt.Sprintf("('message', %d)", timestamp),
+				Table:    "distributed_column_evolution_metadata",
+				Columns:  []string{"signal", "column_name", "column_type", "field_context", "field_name", "version", "release_time"},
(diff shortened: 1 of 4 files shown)


cmd/signozschemamigrator/schema_migrator/manager.go
```diff
@@ -713,30 +713,43 @@ func (m *MigrationManager) executeSyncOperations(ctx context.Context, operations
...
+	}
+
+	// If any of the operations is sync, return false
+	if !item.IsMutation() && item.IsIdempotent() && item.IsLightweight() {
+		return false
+	}
+
+	// If any of the operations is not idempotent, return false
+	if !item.IsIdempotent() {
+		return false
```
srikanthccv: Can you help me with this part, how does operation not idepotent makes it not async?

srikanthccv (commented):
Should the test rather be not mixing the sync and async operations than asserting entire migration record is either sync or async?

therealpandey:
> Should the test rather be not mixing the sync and async operations than asserting entire migration record is either sync or async?

`checkSchemaMigratorMigrationRecords` function in `schema_migrator_test.go` is checking exactly this.

cmd/signozschemamigrator/schema_migrator/metrics_migrations_test.go
```diff
@@ -0,0 +1,30 @@
+package schemamigrator
+
+import (
+	"testing"
+)
+
+func TestMetricsMigrations(t *testing.T) {
+	manager := newTestMigrationManager(t)
+	checkSchemaMigratorMigrationRecords(t, manager, MetricsMigrations)
+}
```
therealpandey: @srikanthccv this is checking whether all operations in a single migration are all sync or all async

PR #797, opened 2026-04-06
feat: add signozllmcostprocessor

nityanandagohain:
### Summary
Adds llm cost processor

### Related Issues
FIxes https://github.com/SigNoz/engineering-pod/issues/4358

### Resource Impact
- [x] Yes
- [ ] No

If **Yes**, describe:


processor/signozllmpricingprocessor/config.go
```diff
@@ -0,0 +1,108 @@
+package signozllmpricingprocessor // import "github.com/SigNoz/signoz-otel-collector/processor/signozllmpricingprocessor"
+
+import (
+	"fmt"
+	"path"
+)
+
+const (
+	CacheModeSubtract = "subtract" // cache_read is already counted inside input_tokens (e.g. OpenAI)
+	CacheModeAdditive = "additive" // cache_read is separate from input_tokens (e.g. Anthropic)
```
piyushsingariya: nit:CacheMode as new type?

processor/signozllmpricingprocessor/config.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	unit := c.DefaultPricing.Unit
+	if unit == "" {
+		unit = UnitPerMillionTokens
+	}
+	if unit != UnitPerMillionTokens {
+		return fmt.Errorf("default_pricing.unit %q is not supported, must be %q", unit, UnitPerMillionTokens)
+	}
+
```
piyushsingariya: the comparison doesn't make sense, I think unit is not supposed to be an input?

processor/signozllmpricingprocessor/processor.go
```diff
@@ -0,0 +1,217 @@
...
+	modelVal, ok := attrs.Get(p.modelAttr)
+	if !ok {
+		return
+	}
+	model := modelVal.Str()
+
+	rule := p.matchRule(model)
+	if rule == nil {
+		return
+	}
```
piyushsingariya, line 115: This will come into action when Span has model but Customer is not observing/evaluating it. right?

processor/signozllmpricingprocessor/config.go
nityanandagohain: Yeah, not really used now. Removing it.

processor/signozllmpricingprocessor/processor.go
nityanandagohain: Yes or we don't have a pricing for the model that the span has.

processor/signozllmpricingprocessor/config.go
nityanandagohain: done

piyushsingariya (dismissed):

piyushsingariya (approved):

PR #811, opened 2026-04-14
feat: add support for scope fields in traces

### LINKED ISSUE #5319: (task) add instrumentation scope to metrics and traces
Add instrumentation scope to metrics and traces.

Related issue https://github.com/SigNoz/signoz/issues/2705 (PR https://github.com/SigNoz/signoz/pull/6339)

krsoninikhil:
### Summary
Add support for instrumentation scope related fields (name, version and attributes) in collector. PR on signoz query builder https://github.com/SigNoz/signoz/pull/10920

image

### Related Issues

Closes https://github.com/SigNoz/signoz/issues/5319

### Resource Impact
- [ ] Yes
- [x] No

If **Yes**, describe:
  

exporter/clickhousetracesexporter/clickhouse_exporter.go
```diff
@@ -101,6 +104,10 @@ const (
 			?,
 			?,
 			?,
+			?,
+			?,
+			?,
+			?,
```
krsoninikhil, line 110: There was one less placeholder in the query. Looks like clickhouse-go handles this automatically so it was not breaking.

krsoninikhil:
Lint errors are not related to change and can be picked up separately.

srikanthccv:
Please use one JSON column instead of three columns and start with 0 max dynamic paths for attributes. It rarely happens that anyone queries on scope attributes (many don't even understand or aware the scope stuff) so start with everything shared. We will materialise when this need to be materialised https://github.com/SigNoz/engineering-pod/issues/4246. 
```
    `scope` JSON(
                name String,
                version String,
                attributes JSON(max_dynamic_paths = 0)) CODEC(ZSTD(1)),

```

krsoninikhil:
After using `scope` columns as json:
image

exporter/clickhousetracesexporter/writer.go
```diff
@@ -160,6 +161,15 @@ func (w *SpanWriter) writeIndexBatchV3(ctx context.Context, batchSpans []*SpanV3
 			w.logger.Warn("resourcemap exceeded the limit of 100 keys")
 		}
 
+		if len(span.Scope.Attributes) > 100 {
+			w.logger.Warn("scope attributes exceeded the limit of 100 keys", zap.String("span_id", span.SpanId))
```
nityanandagohain: let's remove this as there is no actionable for this.

exporter/clickhousetracesexporter/schema-signoz.go
```diff
@@ -207,6 +211,39 @@ type SpanAttribute struct {
 	IsColumn    bool
 }
 
+type InstrumentationScope struct {
+	Name       string            `ch:"name" json:"name"`
+	Version    string            `ch:"version" json:"version"`
+	Attributes map[string]string `ch:"attributes" json:"attributes"`
```
nityanandagohain, line 219: I think it's okay to start with map[string]string as we are not enforcing anything on db and can evolve as needed or should we keep this non restrictive on types ? @srikanthccv

nityanandagohain (commented):
looks good, few comments.

exporter/clickhousetracesexporter/writer.go
krsoninikhil: If there are too many attributes, would be not want to update the limit in future?

nityanandagohain: we should use `JSON(max_dynamic_paths = 0))` right ?

krsoninikhil: Oh right, I think I miss understood this setting. Updated in [b8dfcdc](https://github.com/SigNoz/signoz-otel-collector/pull/811/commits/b8dfcdcc1854316da01ee4dab40146c56cbb9555)

exporter/clickhousetracesexporter/schema-signoz.go
krsoninikhil: Noticed we are converting resource attributes to json [here](https://github.com/SigNoz/signoz-otel-collector/blob/main/exporter/clickhousetracesexporter/clickhouse_exporter_v3.go#L391), followed the similar pattern.

nityanandagohain: Yeah for resources we have just limited to string attributes to start with. Check with @piyushsingariya  once on this if we are good on this without any restriction or we should restrict it.

krsoninikhil:
Added support for number, bool and string data types.
image

nityanandagohain (dismissed):

nityanandagohain:
do a sanity testing by deploying it on one of the staging instances and checking the ingested data.

exporter/clickhousetracesexporter/schema-signoz.go
```diff
@@ -207,6 +211,68 @@ type SpanAttribute struct {
...
+}
+
+func NewInstrumentationScope(scope pcommon.InstrumentationScope) InstrumentationScope {
+	attrs := make(map[string]any)
+	scope.Attributes().Range(func(k string, v pcommon.Value) bool {
+		switch v.Type() {
+		case pcommon.ValueTypeDouble:
+			attrs[k] = v.Double()
+		case pcommon.ValueTypeInt:
+			attrs[k] = float64(v.Int())
```
piyushsingariya: one thing to note is that, even when we're flipping int type to Float, ClickHouse is receiving stringified JSON and it would consider int as int64 only

exporter/clickhousetracesexporter/schema-signoz.go
```diff
@@ -207,6 +211,68 @@ type SpanAttribute struct {
...
+			StringValue: v,
+			IsColumn:    false,
+		})
+	}
+
+	for k, v := range s.Attributes {
+		sa := SpanAttribute{Key: k, TagType: "scope", IsColumn: false}
+		switch val := v.(type) {
+		case float64:
+			sa.DataType = "float64"
```
piyushsingariya: there's already datatype enums better use them directly

krsoninikhil: Are you referring to these? https://github.com/SigNoz/signoz-otel-collector/blob/5914301602cc0976f117478845d9ea187b7d7b86/utils/utils.go#L51

Also, for `SpanAttribute` should we use float64 incase of both float64 and int64?

exporter/clickhousetracesexporter/schema-signoz.go
krsoninikhil: Reverted to saving all values as string in [4117079](https://github.com/SigNoz/signoz-otel-collector/pull/811/commits/4117079bd97ce4049380ba554107eecb8e852715)

nityanandagohain (dismissed):

piyushsingariya (dismissed):

exporter/clickhousetracesexporter/schema-signoz.go
```diff
@@ -207,6 +212,45 @@ type SpanAttribute struct {
...
+
+	spanAttrs := make([]SpanAttribute, 0, len(s.Attributes)+len(scopeFields))
+	for k, v := range scopeFields {
+		if v == "" {
+			continue
+		}
+		spanAttrs = append(spanAttrs, SpanAttribute{
+			Key:         k,
+			TagType:     "scope",
+			DataType:    "string",
```
piyushsingariya: datatype is still referred by, refer ENUM instead

krsoninikhil: Oh missed this, fixed in [be5b95e](https://github.com/SigNoz/signoz-otel-collector/pull/811/commits/be5b95e11704840c6af24aa3eeec5ebbe5bfde65)

piyushsingariya (approved):

exporter/clickhousetracesexporter/schema-signoz.go
```diff
@@ -207,6 +213,45 @@ type SpanAttribute struct {
...
+		spanAttrs = append(spanAttrs, SpanAttribute{
+			Key:         k,
+			TagType:     "scope",
+			DataType:    utils.FieldDataTypeString.String(),
+			StringValue: v,
+			IsColumn:    false,
+		})
+	}
+	return spanAttrs
+}
```
tushar-signoz, line 253: Hey @srikanthccv 
There is an inconsistency here. `scope` prefix is appended in the `name` and `version` keys but is not added for the rest. 
So if we use the de-normalized names, these become
`scope.scope.name`, `scope.scope.version` and `scope.*` etc.
This is not a problem as such but forces us to handle name and version differently.

Added benefit of this is that `scope.name` and `scope.version` will come as first class suggestion in the filter. This also makes the UX for logs and traces consistent. But it goes against the simplification of getting rid of contexts from the signoz UI. That do you suggest here ?

tushar-signoz: Hey @krsoninikhil 
Lets try to be consistent here. The fieldName collision can happen any between intrinsic fields as well as with attributes or resource fields. 
What would be the cost of removing `scope` prefix from `name` and `version` ?

krsoninikhil: There is no cost, just need to decide that a query like `version  = '1.0.0'` when user meant to query scope.version is a valid query or not. If we allow this, then querying for scope.name becomes `name = '--'` which doesn't seem right. `name` and `version` are not attributes, for us, they should act as a field like `scope.name` and `scope.version` similar to how in logs we have `scope_name` and `scope_version`.

tushar-signoz: Hey @krsoninikhil 
Lets go ahead with removing the prefix `scope` from the fields. 
We'll handle the suggestions in the QB.

PR #815, opened 2026-04-30
chore: add telemetry to track processed JSON logs

piyushsingariya:
### Summary

This PR adds two metrics to track count of JSON logs processed in metadata and normalizer processor. Throwing Error when Processor is not there/missing, and Exporters (metadata/clickhouse) both will actually fail in this case, resulting into whole JSON Logs not working.

### Resource Impact
- [ ] Yes
- [x] No

### DIFF
--- a/exporter/clickhouselogsexporter/exporter.go
+++ b/exporter/clickhouselogsexporter/exporter.go
@@ -295,6 +295,7 @@ type clickhouseLogsExporter struct {
 	closeChan chan struct{}
 
 	durationHistogram metric.Float64Histogram
+	jsonLogsProcessed metric.Int64Counter
 
 	keysCache *ttlcache.Cache[string, struct{}]
 	rfCache   *ttlcache.Cache[string, struct{}]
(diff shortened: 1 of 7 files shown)


exporter/clickhouselogsexporter/exporter.go
```diff
@@ -829,25 +832,29 @@ producerIteration:
...
-		// set values to promoted and bodyJSON
-		promoted = buildPromoted(body, promotedSet)
-		// switch the reference to bodyJSON
-		bodyJSON = body
-
-		if !e.bodyJSONOldBodyEnabled {
-			// set body to empty string
-			body = pcommon.NewValueEmpty()
+	if e.bodyJSONEnabled {
+		if body.Type() == pcommon.ValueTypeMap {
```
nityanandagohain, line 839: This is something that is to be caught during a release, i.e during a restart it will make sure that logs are sent to json column. Do you think anywhere something might break during runtime ?

piyushsingariya: Throwing Error when Body is not Map means processor was missing plus both exporter will fail to achieve their tasks, resulting into whole JSON Logs not working.

This result into specific targetting that Ingestion worked fine. and rest of the alerts can be on the side of QB

nityanandagohain (approved):

PR #819, opened 2026-05-14
fix: add normalised fields to the metadata for instrumentation scope

### LINKED ISSUE #5319: (task) add instrumentation scope to metrics and traces

krsoninikhil:
### Summary
Earlier we had added the trace instrumentation scope fields in attribute key table for name and version which included the context as well assuming that we'll restrict the querying to not normalize `scope.version` and `scope.name` to `version` and `name` for readability (see https://github.com/SigNoz/signoz-otel-collector/pull/811). But it seems unnecessary to handle the scope fields differently instead of treating them like every other attribute.

TRD: https://www.notion.so/signoz/Support-for-instrumentation-scope-fields-in-traces-364fcc6bcd1980cba15cd322d5dd7fc9?v=2a7fcc6bcd198049b122000c4ff33372&source=copy_link

### Related Issues

Closes https://github.com/SigNoz/signoz/issues/5319

### Resource Impact
- [ ] Yes
- [x] No

### DIFF
--- a/exporter/clickhousetracesexporter/schema-signoz.go
+++ b/exporter/clickhousetracesexporter/schema-signoz.go
@@ -233,7 +233,7 @@ func NewInstrumentationScope(scope pcommon.InstrumentationScope) Instrumentation
 }
 
 func (s InstrumentationScope) GetSpanAttributes() []SpanAttribute {
-	scopeFields := map[string]string{"scope.name": s.Name, "scope.version": s.Version}
+	scopeFields := map[string]string{"name": s.Name, "version": s.Version}
 	maps.Copy(scopeFields, s.Attributes)
 
 	spanAttrs := make([]SpanAttribute, 0, len(s.Attributes)+len(scopeFields))


ankitnayan:
>But it seems unnecessary to handle the scope fields differently instead of treating them like every other attribute.

@tushar-signoz this will not cause a collision with other attributes with the same name and return un-necessary warnings? Is it not better to be explicit?

tushar-signoz:
> > But it seems unnecessary to handle the scope fields differently instead of treating them like every other attribute.
> 
> @tushar-signoz this will not cause a collision with other attributes with the same name and return un-necessary warnings? Is it not better to be explicit?

Hey @ankitnayan 
So lets assume that there are 3 scope fields, `name`, `version`, `env` . And trace by default has `name` and attributes `version` and `env`
So 
`scope.name`, `scope.version` and `scope.env`
`name` 
`attribute.version` and `attribute.env`

Now consider following filter expressions
`version = '1.0.0'`
`name = 'trace_name'`
`env = 'test'`

What should be sql generated for these ? (Answering this gives a user perspective)

Edit:
With the proposed changes
`version = '1.0.0'` => `scope.verison = '1.0.0' or attribute.version = '1.0.0'`
`name = 'trace_name'` => `name = 'trace_name' or scope.name = 'trace_name'`
`env = 'test'` => `attribute.env = 'test' or scope.env = 'test'`

ankitnayan:
That is the exact confusion and performance degradation I am talking about. If the user wants to search for `scope.version` why unnecessarily make it search in `version` of attributes? UX is explicit searching by `scope.version` and hence should search only for `scope.version` and not `version`

Also, scope is not part of attributes. It is the top level field like trace_id. Is there any limitation that we could not add to top level?

Screenshot 2026-05-18 at 2 18 30 PM

tushar-signoz:
> That is the exact confusion and performance degradation I am talking about. If the user wants to search for `scope.version` why unnecessarily make it search in `version` of attributes? UX is explicit searching by `scope.version` and hence should search only for `scope.version` and not `version`

@ankitnayan 
So if user searches for `scope.version = '1.0.0` then it we will search for exact `scope.version = '1.0.0'`
The question is here if user searches for `version = '1.0.0'` then should we look in both attributes and scope or only attributes. 

> Also, scope is not part of attributes. It is the top level field like trace_id. Is there any limitation that we could not add to top level?

Yes, scope is already a dedicated top level field with type json.

ankitnayan:
ok.. then I am good

tushar-signoz (approved):
LGTM

tushar-signoz:
@krsoninikhil 
Go test is failling, update the test that is checking the name

krsoninikhil:
Closing this based on discussion here -  https://signoz-team.slack.com/archives/C089D1B5516/p1779112987866759?thread_ts=1779090287.211609&cid=C089D1B5516\

MOMs;

If we use name as field name, product will show warning of conflicting context for every query today that expects name to be span name because of something that we shipped. User will have to change all these usage to span.name to avoid this warning. But we can handle this at QB level to new semantics that allow specifying preference of context for some attributes.
Searching attributes across context in case of missing context was decided to cover for user’s mistake of sending the same attribute with multiple contexts. In this case, name has inherent conflict in our system and not a user’s mistake and should be handled by our system itself and not rely on user for resolution.
name and version are not same as attributes, they are intrinsic properties of scope. Tomorrow, any new semantic like Entity can have name property.
We decided to require context prefix in the field name for intrinsic properties of any top level fields like if tomorrow resource gets a new top level property like kind in OTEL, then it should be ingested as resource.kind . This is based on the assumption (with verification) that current QB normalization and any other component doesn’t create any issues because of this decision.

PR #821, opened 2026-05-15
feat(signozlogspipelineprocessor): rebuild ProcessLogs around an async stanza adapter

therealpandey:
**Stacked on #820.** Base = `fix/logspipeline-remove-fanout`. Review #820 first; this PR includes its diff.

## Summary

Rebuilds `signozlogspipelineprocessor` around the same architecture as `opentelemetry-collector-contrib`'s `logstransformprocessor`:

```
ConsumeLogs ──▶ FromPdataConverter (NumCPU workers, parallel plog→stanza)
                       │
                       ▼
              converterLoop (one goroutine, serial pipeline feed)
                       │
                       ▼
              firstOperator.Process(ctx, e) for each entry
                       │
                       ▼
              BatchingLogEmitter (callback)
                       │
                       ▼
              consumeStanzaLogEntries → convertEntriesToPlogs → next consumer
```

`ConsumeLogs` enqueues into `FromPdataConverter.Batch` and returns immediately. The factory bypasses `processorhelper.NewLogs` (which only supports sync `ProcessLogs` callbacks) and returns the processor directly as a `processor.Logs`.

## What you give up (deliberate)

This is the **opposite** of the original design intent of this processor. The pre-existing doc comment (now removed) literally said *"A sync implementation … logstransform processor in opentelemetry-collector-contrib is async."* By going async we lose:

1. **Backpressure to the OTel batchprocessor.** `ConsumeLogs` returns nil immediately. The batchprocessor sees every batch as instantly successful and will keep pulling Kafka regardless of how slow the pipeline is. Memory grows in `FromPdataConverter.workerChan` and the BatchingLogEmitter buffer instead of being throttled upstream.
2. **Error propagation from the next consumer.** If `clickhouselogsexporter.ConsumeLogs` returns an error, it's logged in `consumeStanzaLogEntries` and dropped on the floor. The receiver thinks everything succeeded.
3. **`otelcol_processor_signozlogspipeline_*` latency histograms change meaning** — they now measure "time to enqueue to FromConverter", not end-to-end pipeline time.

## What you get

`plogToEntries` now happens in parallel across `NumCPU` `FromPdataConverter` workers. On the profile we captured (logs-on, v0.129.6), the conversion step was **~3.4% of total CPU**. This change should claw back most of that. The stanza pipeline operators themselves still run serially in `converterLoop` — same as contrib, same as #820 left them. No additional parallelism at the pipeline stage.

## Behavior preservation — router unmatched-entry pass-through

The original sync implementation silently passed unmatched-by-router entries through to the next consumer (because `convertEntriesToPlogs` walked the full input slice regardless of pipeline forwarding). Contrib's `logstransformprocessor` would drop those entries. **This PR preserves the SigNoz behavior** rather than the contrib behavior:

- `router.Transformer` gains a `fallthroughOutput operator.Operator` field and a `SetFallthroughOutput(op)` setter.
- The processor wires each router's fallthroughOutput to the `BatchingLogEmitter` after pipeline build (`processor.go:startPipeline`).
- When no route matches, the router forwards the entry directly to the emitter (skipping any downstream operators — which is exactly what the old code did, since downstream operators were unreachable for unmatched entries in the sync model).
- `router/body_not_json` test asserts this preservation and passes through the standard table-driven test loop.

## Compatibility notes

- **`InternalTempAttributePrefix` filtering preserved.** The emitter callback uses the local `convertEntriesToPlogs` (utils.go) rather than `adapter.ConvertEntries` so SigNoz-specific scratch attributes are still filtered out of the output.
- **`MutatesData: true`** still declared via the processor's `Capabilities()` method.
- The `processorCapabilities` package var in `factory.go` was removed — capabilities now come from the processor's own method.

## Test changes

- `validateProcessorBehavior` and `TestConcurrentConsumeLogs` wait for the sink with `require.Eventually` (was: read sink immediately after `ConsumeLogs` returned).
- `TestConcurrentConsumeLogs` asserts on **log record count** rather than number of plog batches, because `BatchingLogEmitter` merges entries into fewer batches.
- Every test now `defer proc.Shutdown(ctx)` to drain `FromPdataConverter` worker goroutines — otherwise the race detector sees them as leaks across the package.

## Test plan

- [x] `go test ./processor/signozlogspipelineprocessor/... -race -count=1` passes
- [x] `go build ./processor/signozlogspipelineprocessor/` clean
- [x] router/body_not_json verifies unmatched-entry pass-through is preserved end-to-end
- [ ] Staging deploy: capture CPU profile, confirm `plogToEntries`-side work is parallelised across `NumCPU` `FromPdataConverter.workerLoop` goroutines
- [ ] Staging: confirm `otelcol_exporter_send_failed_log_records` is being monitored, since exporter errors no longer back-propagate through the processor

### DIFF
--- a/processor/signozlogspipelineprocessor/factory.go
+++ b/processor/signozlogspipelineprocessor/factory.go
@@ -1,5 +1,7 @@
-// Brought in as is from logstransform processor in opentelemetry-collector-contrib
-// with identifiers changed for the new processor
+// Async processor: ConsumeLogs returns immediately after enqueuing to
+// FromPdataConverter. The factory therefore bypasses processorhelper.NewLogs
+// (which only supports sync ProcessLogs callbacks) and returns the processor
+// directly as a processor.Logs implementation.
 package signozlogspipelineprocessor
 
 import (
@@ -12,7 +14,6 @@ import (
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/processor"
-	"go.opentelemetry.io/collector/processor/processorhelper"
 
 	"github.com/SigNoz/signoz-otel-collector/processor/signozlogspipelineprocessor/internal/metadata"
 )
@@ -33,10 +34,8 @@ func createDefaultConfig() component.Config {
 	}
 }
 
-var processorCapabilities = consumer.Capabilities{MutatesData: true}
-
 func createLogsProcessor(
-	ctx context.Context,
+	_ context.Context,
 	set processor.Settings,
 	cfg component.Config,
 	nextConsumer consumer.Logs,
@@ -49,19 +48,10 @@ func createLogsProcessor(
 		return nil, errors.New("no operators were configured for signozlogspipeline processor")
 	}
 
-	proc, err := newLogsPipelineProcessor(pCfg, set.TelemetrySettings)
+	proc, err := newLogsPipelineProcessor(pCfg, set.TelemetrySettings, nextConsumer)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't build \"signozlogspipeline\" processor %w", err)
 	}
 
-	return processorhelper.NewLogs(
-		ctx,
-		set,
-		cfg,
-		nextConsumer,
-		proc.ProcessLogs,
-		processorhelper.WithStart(proc.Start),
-		processorhelper.WithShutdown(proc.Shutdown),
-		processorhelper.WithCapabilities(processorCapabilities),
-	)
+	return proc, nil
 }
--- a/processor/signozlogspipelineprocessor/stanza/operator/operators/router/transformer.go
+++ b/processor/signozlogspipelineprocessor/stanza/operator/operators/router/transformer.go
@@ -22,6 +22,20 @@ type Transformer struct {
 	helper.BasicOperator
 	routes              []*Route
 	allCompiledPatterns map[string]func(s string) bool
+
+	// fallthroughOutput receives entries that no route matched. When nil, those
+	// entries are dropped — matching contrib's logstransformprocessor behavior.
+	// The SigNoz processor sets this to the pipeline's BatchingLogEmitter so
+	// unmatched entries flow through to the next consumer unchanged, preserving
+	// the behavior of the pre-async sync implementation (where unmatched
+	// entries remained in the input slice and were returned as-is).
+	fallthroughOutput operator.Operator
+}
+
+// SetFallthroughOutput wires the operator that receives entries no route
+// matched. Pass nil to drop unmatched entries.
+func (t *Transformer) SetFallthroughOutput(op operator.Operator) {
+	t.fallthroughOutput = op
 }
 
 // Route is a route on a router operator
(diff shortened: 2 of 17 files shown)


processor/signozlogspipelineprocessor/factory.go
```diff
@@ -49,19 +48,10 @@ func createLogsProcessor(
...
-	return processorhelper.NewLogs(
-		ctx,
-		set,
-		cfg,
-		nextConsumer,
-		proc.ProcessLogs,
-		processorhelper.WithStart(proc.Start),
-		processorhelper.WithShutdown(proc.Shutdown),
-		processorhelper.WithCapabilities(processorCapabilities),
-	)
```
piyushsingariya, line 66: nit: Why do we need to move away from processorhelper?

processor/signozlogspipelineprocessor/stanza/operator/operators/router/transformer.go
```diff
@@ -22,6 +22,20 @@ type Transformer struct {
...
 	routes              []*Route
 	allCompiledPatterns map[string]func(s string) bool
+
+	// fallthroughOutput receives entries that no route matched. When nil, those
+	// entries are dropped — matching contrib's logstransformprocessor behavior.
+	// The SigNoz processor sets this to the pipeline's BatchingLogEmitter so
+	// unmatched entries flow through to the next consumer unchanged, preserving
+	// the behavior of the pre-async sync implementation (where unmatched
+	// entries remained in the input slice and were returned as-is).
+	fallthroughOutput operator.Operator
```
piyushsingariya: I've a doubt here, I'll research myself too, Why this field is necessary. If no routes match, unmatched entries still pass to Next processor? like OTEL Collector should do that by itself.

piyushsingariya (commented):
pulling it to local and checking

processor/signozlogspipelineprocessor/stanza/operator/operators/router/transformer.go
piyushsingariya:
```suggestion
	defaultOutput operator.Operator
```

processor/signozlogspipelineprocessor/stanza/operator/operators/router/transformer.go
```diff
@@ -22,6 +22,20 @@ type Transformer struct {
...
+}
+
+// SetFallthroughOutput wires the operator that receives entries no route
+// matched. Pass nil to drop unmatched entries.
+func (t *Transformer) SetFallthroughOutput(op operator.Operator) {
```
piyushsingariya:
```suggestion
func (t *Transformer) SetDefaultOutput(op operator.Operator) {
```

processor/signozlogspipelineprocessor/stanza/operator/operators/router/transformer.go
```diff
@@ -63,9 +79,15 @@ func (t *Transformer) Process(ctx context.Context, entry *entry.Entry) error {
...
-				break
+				return nil
 			}
 		}
+
+		// No route matched. Preserve the unmatched entry by forwarding it to
+		// the fallthrough output if one is configured.
+		if t.fallthroughOutput != nil {
+			_ = t.fallthroughOutput.Process(ctx, entry)
+		}
```
piyushsingariya: defaultOutput can not be nil. Hard check required

piyushsingariya (dismissed):
deafultOutput aligns with Stanza naming, Overall I've verified this should work. Sanity testing still required.

piyushsingariya:
I've tested these cases

1. An entry being saved not matching Pipeline's filter
2. An entry matching pipeline's filter but not matching one of the steps
3. An entry matching pipeline's filter plus all the steps too

nityanandagohain (dismissed):
changes looks good to me, let get it running on staging forsometime, we will collect the telemetry and then release it.

therealpandey:
@piyushsingariya one flaky test in logspipeline.

piyushsingariya:
> @piyushsingariya one flaky test in logspipeline.

Actually the test needed an upgrade, now that it's async, these logs were getting rejected and eventually was failing, I needed to add a default in router, as our SigNoz does when sending yaml over OpAMP

nityanandagohain (approved):
lets release it more cautiously.

PR #822, opened 2026-05-15
fix: remove unnecessary json.Marshal and Iteration.

nityanandagohain:
### Summary
emove unnecessary json.Marshal and Iteration.

### Resource Impact
- [X] Yes
- [ ] No

If **Yes**, describe:
It will reduce resource usage


nityanandagohain:
Can be kept as draft as metering is not fully rolled out.

PR #824, opened 2026-05-27
chore: Update promoted tests

piyushsingariya:
### Summary

Update tests for better sanity about two cases
1. Where parent is updated 
2. Where leaf node is Array

## Resource Impact
- [ ] Yes
- [x] No

If **Yes**, describe:
  

exporter/clickhouselogsexporter/promoted_test.go
```diff
@@ -53,6 +53,47 @@ func TestPromotedPathSeparation(t *testing.T) {
...
+				"user": {},
+			},
+			expectedPromoted: map[string]interface{}{},
+		},
+		{
+			name: "array_leaf_found_promoted",
+			body: map[string]interface{}{
+				"message": "test log",
+				"user": map[string]interface{}{
+					"orders": []any{
```
nityanandagohain, line 76: so we allow array nodes to be promoted ?

piyushsingariya: the leaf itself is an array. Any path inside `orders` can not be promoted

nityanandagohain: If this array has multiple nested structs inside it that will be inside same column or clickhouse will create more columns internally ?

piyushsingariya: It's not officially documented.

piyushsingariya: But I think it does, cuz max_dynamic_paths have some value at every level, and it keeps dividing by 4 at each level

nityanandagohain: then this might create unnecessary columns if the array is bad.

Do we allow promoting a path inside array ?

piyushsingariya: No. It's complicated to achieve

nityanandagohain: Then let's start with not allowing this as well and document it ?

piyushsingariya: I don't think we should. It's simpler with "**promoted paths regardless of type gets promoted**" It probably/definitely creates more internal columns with arrays on disk, but it's not part of the 1024 dynamic paths at the top level.

nityanandagohain: >It probably creates more internal columns with arrays on disk, but it's not part of the 1024 dynamic paths at the top level

I think we should confirm this before taking a decision here ?

piyushsingariya: I'll raise a question ticket on ClickHouse

nityanandagohain: If you just ingest some dummy data you will be able to find out ?

piyushsingariya: Yes, I've done that. That's how we know `max_dynamic_paths` keeps getting divided by 4 at each level
1024 -> 256 -> 64 -> 16 -> 0 and `max_dynamic_types` gets divided by 2; 32 -> 16 -> 8 -> 4 -> 2
This is how ClickHouse handles that they unintentionally don't bloat disk due to arrays, the budget keeps decreasing.

And even if customer ingests bad array once, ClickHouse has a logic that prioritizes high occurrence paths as Dynamic Ones and degrade others as shared paths at those levels.

So it is not a problem. If a customer is ingesting such deeply nested data, they need to rethink their JSON Schema tbh. Let's get on a call?

nityanandagohain (approved):

PR #828, opened 2026-05-29
fix: don't pass signoz collector ID into spans

namanv3:
### Summary
Spans currently carry `signoz.collector.id` as a resource attribute, which is happening as a side effect of adding this attribute to metrics. This is not needed in spans, and can in fact cause slowness when querying.

The issue was because when this attribute was being added to metrics, the same map was being used for traces, so traces got this flag as well. To fix this, metrics will get a copy of the traces map and then only have this attribute added.

### Some screenshots post this change

An example span after the change was made, with it not having this resource attribute anymore.
Screenshot 2026-05-29 at 17 43 35

No spans have this attribute after the change was made, and all spans before this change had this attribute. 
Screenshot 2026-05-29 at 17 43 42

Metrics still carry this attribute (different ID because collector restarted with a different ID).
Screenshot 2026-05-29 at 17 43 58

### Related Issues

Closes https://github.com/SigNoz/engineering-pod/issues/5136#event-26094817898

### Resource Impact
- [ ] Yes
- [x] No


processor/signozspanmetricsprocessor/processor.go
```diff
@@ -1037,15 +1037,19 @@ func (p *processorImp) aggregateMetrics(traces ptrace.Traces) {
 		if !ok {
 			continue
 		}
-		resourceAttr.PutStr(signozID, p.instanceID)
+		// Stamp signoz.collector.id on a copy used only for metrics, so the
+		// forwarded spans stay unmodified.
+		metricResourceAttr := pcommon.NewMap()
+		resourceAttr.CopyTo(metricResourceAttr)
+		metricResourceAttr.PutStr(signozID, p.instanceID)
```
srikanthccv: let's avoid copying here. we should be able to add this to key without it being part of the map.

therealpandey:
Pushing this as part of https://github.com/SigNoz/platform-pod/issues/2343

processor/signozspanmetricsprocessor/processor.go
namanv3: @srikanthccv I've removed the map copying. I haven't put in the collector ID in the metricKey however, because this key is never deconstructed to fetch the dimensions. Only the caches have the value and the methods that fill that in have been updated.

I've also kept the label set for metrics the same.

srikanthccv (approved):

srikanthccv:
verified locally and patient-bat

PR #833, opened 2026-06-03
feat(trace-json): store trace attributes as json

krsoninikhil:
### Summary
Store attributes column of json type to store the attributes as raw json. Not adding any index in values for now since `toString(attributes)` is expensive and `JSONAllValues` is only supported in Clickhouse v26.4

Breaking Change: This change requires clickhouse to be atleast v25.12.5

### Related Issues

Part of https://github.com/SigNoz/engineering-pod/issues/5068

### Resource Impact
- [ ] Yes
- [x] No

### DIFF
--- a/cmd/signozschemamigrator/schema_migrator/traces_migrations.go
+++ b/cmd/signozschemamigrator/schema_migrator/traces_migrations.go
@@ -1184,8 +1184,23 @@ var TracesMigrations = []SchemaMigrationRecord{
 				},
 				After: &Column{Name: "attributes_bool"},
 			},
+			AlterTableAddIndex{
+				Database: "signoz_traces",
+				Table:    "signoz_index_v3",
+				Index: Index{
+					Name:        "attributes_paths_tokenbf",
+					Expression:  JSONPathsIndexExpr("attributes"),
+					Type:        "tokenbf_v1(1024, 2, 0)",
+					Granularity: 1,
+				},
+			},
 		},
 		DownItems: []Operation{
+			AlterTableDropIndex{
+				Database: "signoz_traces",
+				Table:    "signoz_index_v3",
+				Index:    Index{Name: "attributes_paths_tokenbf"},
+			},
 			AlterTableDropColumn{
 				Database: "signoz_traces",
 				Table:    "distributed_signoz_index_v3",
--- a/cmd/signozschemamigrator/schema_migrator/traces_migrations.go
+++ b/cmd/signozschemamigrator/schema_migrator/traces_migrations.go
@@ -1161,4 +1161,71 @@ var TracesMigrations = []SchemaMigrationRecord{
 			},
 		},
 	},
+	{
+		MigrationID: 1011,
+		UpItems: []Operation{
+			AlterTableAddColumn{
+				Database: "signoz_traces",
+				Table:    "signoz_index_v3",
+				Column: Column{
+					Name:  "attributes",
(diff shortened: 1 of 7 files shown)


exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -239,6 +239,20 @@ func (attrMap *attributesData) add(key string, value pcommon.Value) {
...
+		result[k] = v
+	}
+	for k, v := range attrMap.NumberMap {
+		result[k] = v
+	}
+	for k, v := range attrMap.BoolMap {
+		result[k] = v
+	}
+	return result
+}
```
krsoninikhil: Incase we are planning to drop type specific maps in future and don't want dependency on it, we can store the attributes as raw in a new field in `attributesData` like `attributesData.RawMap`.

cmd/signozschemamigrator/schema_migrator/traces_migrations.go
```diff
@@ -1161,4 +1161,56 @@ var TracesMigrations = []SchemaMigrationRecord{
...
+			AlterTableAddIndex{
+				Database: "signoz_traces",
+				Table:    "signoz_index_v3",
+				Index: Index{
+					Name:        "attributes_paths_tokenbf",
+					Expression:  JSONPathsIndexExpr("attributes"),
+					Type:        "tokenbf_v1(1024, 2, 0)",
+					Granularity: 1,
+				},
+			},
```
krsoninikhil, line 1227: Only adding the index on path for now, until we plan the clickhouse upgrade to support `JSONAllValues`.

exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
krsoninikhil: Also, do we want to flatten the nested keys or keep them nested?

cmd/signozschemamigrator/schema_migrator/traces_migrations.go
```diff
@@ -1161,4 +1161,64 @@ var TracesMigrations = []SchemaMigrationRecord{
...
 		},
 	},
+	{
+		MigrationID: 1011,
+		UpItems: []Operation{
+			AlterTableModifySettings{
+				Database: "signoz_traces",
+				Table:    "signoz_index_v3",
+				Settings: TableSettings{
+					{Name: "object_serialization_version", Value: "'v3'"},
```
therealpandey, line 1171: What are these?

cmd/signozschemamigrator/schema_migrator/traces_migrations.go
```diff
@@ -1161,4 +1161,77 @@ var TracesMigrations = []SchemaMigrationRecord{
...
+					{Name: "object_shared_data_serialization_version", Value: "'advanced'"},
+				},
+			},
+		},
+		DownItems: []Operation{},
```
krsoninikhil, line 1176: Should we have an operation like `AlterTableResetSettings` for down item on this?

therealpandey (changes requested):

cmd/signozschemamigrator/schema_migrator/traces_migrations.go
therealpandey: Ignore this now, the comment was stuck in pending for the last 3 weeks.

therealpandey (approved):

PR #835, opened 2026-06-04
feat(trace-json): add support for promoted json atttributes for traces

krsoninikhil:
### Summary
Follow up on https://github.com/SigNoz/signoz-otel-collector/pull/833

Add mechanism to sync and save promoted trace attribute paths. Exactly same as how body_v2 json is doing it.

### Related Issues

Closes https://github.com/SigNoz/engineering-pod/issues/5068


exporter/clickhousetracesexporter/promoted.go
```diff
@@ -0,0 +1,42 @@
...
+
+	for _, path := range promotedPaths {
+		handleSingleAttributePath(attrs, pm, path, path)
+	}
+	return pm
+}
+
+// handleSingleAttributePath walks attrs according to remainingPath and extracts the value into promotedMap at fullPath.
+func handleSingleAttributePath(bodyMap pcommon.Map, promotedMap pcommon.Map, fullPath string, remainingPath string) {
+	if v, ok := bodyMap.Get(remainingPath); ok {
```
nityanandagohain: we can use a better naming here instead of `bodyMap` or better, if we can move these logic out of these exporters and keep it common?

exporter/clickhousetracesexporter/writer.go
```diff
@@ -140,6 +150,50 @@ func (e *SpanWriter) fetchShouldSkipKeys() {
...
+	rows := []struct {
+		FieldName string `ch:"field_name"`
+	}{}
+	if err := e.db.Select(context.Background(), &rows, query); err != nil {
+		e.logger.Error("error while fetching promoted attribute paths", zap.Error(err))
+		return
+	}
+
+	// deduplicate
+	seen := make(map[string]struct{}, len(rows))
```
nityanandagohain: instead of dedup here , fetch from clickhouse using FINAL in the query ?

nityanandagohain (commented):
some comments

exporter/clickhousetracesexporter/writer.go
krsoninikhil: It triggers the merge even if the table is already merged. Avoided that based on similar change once we did tried for trace_summary table if you remember where we had to use group by instead. 

Should we try group by here as well?

nityanandagohain: ah ok, yes lets check if group by works. Else we can keep it

exporter/clickhousetracesexporter/promoted.go
krsoninikhil: > we can use a better naming here

Ah, copied from logs. Fixed in [038e282](https://github.com/SigNoz/signoz-otel-collector/pull/835/commits/038e2829d72472d09ec98c64a0c6a7ba501eced0)

> we can move these logic out of these exporters

That would be better, just not sure if we these can evolve independently. Let me move it though.

exporter/clickhousetracesexporter/writer.go
krsoninikhil: Added group by, haven't changed in logs in this PR though.

exporter/clickhousetracesexporter/promoted.go
krsoninikhil: Moved to a util method, had to change the signature to take promoted paths as map though instead of slice to since logs was using that type. 

Tried changing the logs promoted paths to `[]slice` in [25fb614](https://github.com/SigNoz/signoz-otel-collector/pull/835/commits/25fb61457b4bf7b9a3aea4ba5066d3df79ca6c85) but looked like out of scope change for this PR so reverted it. 

I can change this data type when we do json attribute for logs if that is required.

nityanandagohain (approved):

PR #839, opened 2026-06-12
feat: metrics cardinality control - collector implementation

srikanthccv:
## Summary

Fixes https://github.com/SigNoz/engineering-pod/issues/5339

Collector-side implementation of metrics cardinality control: per-metric label-drop rules aggregate away unwanted series at rest in ClickHouse, cutting series count, cost, and churn while preserving correctness (single-writer principle, reset-aware counters).

Tracking: SigNoz/engineering-pod#5264 (epic) / SigNoz/engineering-pod#5339 (collector task — full implementation plan lives there).

> **Delivery note (as agreed):** development lands in this one PR; after the scale test on the internal environment passes, it gets broken down into reviewable PRs that all merge the same day. The commits are phase-aligned so the breakdown is a contiguous-commit-range split.

How it fits together:

```
exporter ─► samples_v4_buffer ───MV(rf=0)──────────► samples_v4 ─► agg_5m/30m
                  └─ refreshable MVs (1m) ─────────► samples_v4_reduced_{last,sum}_60s
exporter ─► time_series_v4_buffer ─MV(raw, rf=0)───► time_series_v4 ─► 6hrs/1day/1week
                  └─ MV(is_reduced) ───────────────► time_series_v4_reduced
```

- The exporter polls `metric_reduction_rules` (written by query-service), computes a `reduced_fingerprint` per datapoint by re-hashing the resource→scope→point chain minus the dropped keys, and lands everything in 1-day-TTL buffer tables. Rules apply by **datapoint timestamp vs the rule's `effective_from`** epoch, so every replica starts reducing at the same data-time boundary regardless of poll timing — no 60s bucket is ever partially reduced.
- Buffer tables shard by `if(reduced_fingerprint != 0, reduced_fingerprint, fingerprint)` so a reduced group's series co-locate on one shard for the per-shard refreshable MVs; unruled rows shard exactly as today and flow to the long-retention tables through incremental feed MVs (cascading into the existing agg/rollup MVs).
- Three refreshable MVs (EMPTY, APPEND, 1m refresh, 10s randomize) aggregate ruled series into 60s reduced tables: gauge/UpDownCounter last+min+max+sum_values, delta sums, and cumulative counters via per-point reset-aware lag. Reads dedup recomputed buckets with `argMax(computed_at)`; `ReplacingMergeTree(computed_at)` keeps the latest recompute on merge.
- Fail-open everywhere: poll failure keeps the last ruleset; anything without an applicable rule is written unreduced, which is always correct, just at full fidelity.

## Commits (phase-aligned)

1. `fix(signozclickhousemetrics)`: emit gauge semantics for synthetic series — `.quantile` → Unspecified + non-monotonic, histogram/exp-hist `.min`/`.max` → non-monotonic (standalone precursor, independent of reduction)
2. `feat(schemamigrator)`: `CreateRefreshableMaterializedViewOperation` (REFRESH EVERY / RANDOMIZE FOR / APPEND / EMPTY / DEFINER) + `ReplacingMergeTree` version-column support
3. `feat(schemamigrator)`: migration 1008 — buffer tables, reduced tables, `metric_reduction_rules`, 3 incremental feed MVs, 3 refreshable MVs, `minmax` skip index on `unix_milli`, down-migrations
4. `feat(fingerprint)`: `Reduced(offset, drop)` — drop-set-aware rehash preserving the chain so series differing only in dropped keys collapse to one reduced fingerprint
5. `feat(signozclickhousemetrics)`: reduction config (default **off**), rules poller, buffer write path, reduced series rows, exp-hist exclusion

## Testing done

- **End-to-end MV semantics against a real ClickHouse** (`clickhouse local`): created the full migration-1008 schema, inserted fixtures, ran `SYSTEM REFRESH VIEW`, asserted outputs:
  - the design doc's Pod A/B/C counter-reset worked example yields exactly **650** (mid-bucket reset handled by the per-point lag);
  - delta sums, gauge last/min/max/sum_values, and series/sample counts correct;
  - cumulative non-monotonic (UpDownCounter) routes to the last-value view;
  - staleness-flagged samples (`bitAnd(flags,1)`) excluded;
  - feed MV passes only `reduced_fingerprint = 0` rows to `samples_v4`.
- **Unit tests**: reduced-fingerprint determinism/collapse across attribute levels (incl. a manual chain cross-validation of the exporter's output), `effective_from` epoch boundary behavior, cross-resource collapse, derived-series rule inheritance (`.bucket` keeps `le`), exp-hist exclusion, config validation, buffer-path `writeBatch` against a mocked connection, golden `ToSQL()` tests for the new DSL operations.
- **Full repo**: `go build ./...`, `go test -count=1 -race ./...` green (one unrelated flake: opamp test binding port 8888, passes on re-run), `go vet` + `gofmt` clean. Local `make lint` is pre-broken on `main` (no `.golangci.yml`, stale install path); relying on CI's pinned linter.
- **Benchmarks**: worst case with **every** metric ruled (10k metrics × 10 points × 30 attrs): ~25–30% prepareBatch time, +27% allocs vs baseline (`Benchmark_prepareBatchGaugeWithReduction`). Overhead scales with the ruled fraction, so blended real-world cost is a few percent. Reduced-series-row construction is gated to once per series per batch to get there.

## Findings during implementation

- **The design doc's gauge MV SQL has a latent bug**: aliasing the 60s bucket as `unix_milli` inside the subquery makes `argMax(value, unix_milli)` resolve to the *alias* (ClickHouse alias precedence over column names), silently degrading last-value selection to an arbitrary value. The shipped MV names the alias `bucket_unix_milli`; the clickhouse-local validation proves the fix.
- **The doc's cumulative MV undercounts intra-bucket resets** (bucket-level `max` + cross-bucket lag gives 150 for Pod A, prose derives 300). Implemented per-point reset-aware lag over raw points, attributing each increment to the later point's bucket so windows tile exactly.
- **DEFINER/SQL SECURITY omitted** from the shipped refreshable MVs: migrations are shared with self-host where the `admin` user may not exist. The DSL supports the clause when cloud wants it.
- Buffer tables named `samples_v4_buffer` / `time_series_v4_buffer` (duration dropped from the name so TTL stays tunable). Effective retention with 1-day TTL + daily partitions + `ttl_only_drop_parts`: 24–48h.
- **Fingerprint changes shipped knowingly** (per the agreed plan): `.quantile` series get new fingerprints (`__temporality__` is part of identity) — querier must tolerate both temporalities for the retention overlap (SigNoz/engineering-pod#5341 contract); exp-hist `.min`/`.max` previously stamped the parent's `Delta` in `__temporality__` while the column said `Unspecified` — fixed for consistency, changes those series' fingerprints (exp-hist is default-off).
- A series whose rule drops only labels it doesn't have is its own reduction; the ts buffer's sort key includes `is_reduced` and the hourly write-cache keys reduced rows separately so both rows survive.

## What's NOT done (intentionally)

- **Phase 5 meter/billing integration** (`signozmeterconnector` dimension split, SigNoz/engineering-pod#5343) — D3 is still a proposal; done elsewhere.
- **Scale test on the internal environment** (SigNoz/engineering-pod#5345 gate) — refresh duration/bytes scanned vs series count and ruled fraction, including the accepted ~2× replica refresh duplication. The validation SQL pattern used locally is reproducible from migration 1008.
- **Querier routing, rules-management API, UI, docs** — other repos (SigNoz/engineering-pod#5340, #5341, #5342, #5344).
- Exp-hist sketches and `metadata`-table paths are untouched by design; reduction is default-off so default deployments behave exactly as today (write targets unchanged until `reduction.enabled: true`).


srikanthccv:
## Bug + fix in the gauge/delta reduction MVs: they slice buckets at the window's lower edge

While scale-testing reduction on a internal cluster, gauge reduction was undercounting badly — reduced gauge values were ~27% of truth, while counters were ~100%. Root cause is a windowing bug in `samples_v4_reduced_last_60s_mv` (and the same gap in `samples_v4_reduced_sum_60s_delta_mv`). Full analysis: SigNoz/engineering-pod#5264.

### The bug

These MVs refresh every minute and re-read a sliding event-time window `[now−5m, now−2m]`, writing each result with `computed_at = now()`. The output is `ReplacingMergeTree(computed_at)` and reads dedup with `argMax(…, computed_at)`, so **the newest `computed_at` per (series, bucket) wins**.

A bucket is fixed in event-time but the window slides forward. The gauge/delta MVs filter and emit on the **raw-sample lower bound only** — there is no "emit only whole buckets" guard. So as `now−5m` recedes *through* a bucket on its way out, the MV keeps re-emitting that bucket **minus its chopped-off front** — a shrinking sliver — each with a newer `computed_at`. `argMax(computed_at)` then serves the **last = most-truncated** sliver.

Live proof on bucket `11:47:00` (one reduced_fp): a complete version `264 samples` @ `computed_at 11:53:41`, then a sliver `18 samples` @ `11:57:56`; the sliver wins. `18/264 = 0.068` = exactly the last `4s/60s` of the bucket still above the receding floor. Aggregate: the served versions held only **33%** of the samples the MV itself had computed; 80k buckets were serving a ≥2× smaller version than one already written.

The **cumulative-counter MV in this same file is already correct** — it reads one minute deeper than it emits and guards on the bucket:
```sql
-- samples_v4_reduced_sum_60s_cumulative_mv (correct)
    AND (unix_milli >= (toUnixTimestamp(now() - toIntervalMinute(6)) * 1000))   -- read [now-6m, now-2m]
  ...
  WHERE (isNaN(increment) = 0)
    AND (bucket_unix_milli >= (toUnixTimestamp(now() - toIntervalMinute(5)) * 1000))   -- emit only buckets <= 5m old
```
That guard is why counters measured ~100%. The gauge and delta MVs are missing it.

### The fix — give the gauge MV the same guard (before → after)

```sql
-- BEFORE: samples_v4_reduced_last_60s_mv (slices on the way out)
            FROM signoz_metrics.samples_v4_buffer
            PREWHERE reduced_fingerprint != 0
            WHERE (bitAnd(flags, 1) = 0)
                AND ((temporality = 'Unspecified') OR ((temporality = 'Cumulative') AND (is_monotonic = false)))
                AND (unix_milli < (toUnixTimestamp(now() - toIntervalSecond(120)) * 1000))
                AND (unix_milli >= (toUnixTimestamp(now() - toIntervalMinute(5)) * 1000))
            GROUP BY env, temporality, metric_name, reduced_fingerprint, fingerprint, bucket_unix_milli
        )
        GROUP BY env, temporality, metric_name, reduced_fingerprint, bucket_unix_milli
```
```sql
-- AFTER: read one minute deeper, emit only whole buckets (never a sliver)
            FROM signoz_metrics.samples_v4_buffer
            PREWHERE reduced_fingerprint != 0
            WHERE (bitAnd(flags, 1) = 0)
                AND ((temporality = 'Unspecified') OR ((temporality = 'Cumulative') AND (is_monotonic = false)))
                AND (unix_milli < (toUnixTimestamp(now() - toIntervalSecond(120)) * 1000))
                AND (unix_milli >= (toUnixTimestamp(now() - toIntervalMinute(11)) * 1000))   -- (5 -> 11) read 1 min below the guard
            GROUP BY env, temporality, metric_name, reduced_fingerprint, fingerprint, bucket_unix_milli
        )
        WHERE bucket_unix_milli >= (toUnixTimestamp(now() - toIntervalMinute(10)) * 1000)    -- NEW: emit only whole buckets
        GROUP BY env, temporality, metric_name, reduced_fingerprint, bucket_unix_milli
```

Two independent things in that diff:
1. **The fix (required):** the added `WHERE bucket_unix_milli >= now−Nm` guard + reading to `now−(N+1)m`. This makes the MV drop a bucket cleanly once it ages past `N` instead of slicing it, so the last/winning emission is always complete. (Same shape as the cumulative MV — there it's `N=5`.)
2. **The lookback `N` (tunable):** how long to keep recomputing a bucket before finalizing it = how much of the late-write tail it captures. I validated `N=10` (shown above); the cumulative MV uses `N=5`. Larger `N` = more complete but later finalization (bucket settles at age `N`).

`sum_60s_delta_mv` has the **identical missing guard** (raw `[now−5m, now−2m]`, no `bucket_unix_milli` filter) and needs the same change — it will undercount delta sums on exiting buckets exactly the same way.

### Validated on cluster (heavy gauge ingestion, `v0.144.5-3c8d618`)

| measurement | before (no guard) | after (guard, N=10) |
|---|---|---|
| served-vs-most-complete (age 11–20) | **0.33** | **1.0000** |
| buckets serving a smaller-than-available version | 80,048 | **0** |
| coverage by bucket age | ~100% in-window → **6%/50% cliff** on exit | **flat 1.00** (age 7–24) |
| aggregate coverage, finalized band (age 11–18) | **0.27** | **1.0000** |

0 refresh exceptions; refresh stayed ~2s (the ~9-min read is well under the 1-min interval). Applied live via `ALTER TABLE … MODIFY QUERY ON CLUSTER`.

cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
```diff
@@ -975,4 +975,679 @@ var MetricsMigrations = []SchemaMigrationRecord{
...
+				},
+				Indexes: []Index{
+					// minmax index lets each refresh prune to the small fresh parts
+					{Name: "idx_unix_milli", Expression: "unix_milli", Type: "minmax", Granularity: 1},
+				},
+				Engine: MergeTree{
+					PartitionBy: "toDate(unix_milli / 1000)",
+					OrderBy:     "(env, temporality, metric_name, fingerprint, unix_milli)",
+					// with ttl_only_drop_parts and daily partitions effective retention is 24-48h
+					TTL: "toDateTime(unix_milli / 1000) + toIntervalSecond(86400)",
```
srikanthccv: alhough the doc had 6h in proposal, changed in to 24h because it gives generous room to make corrections if any till this becomes GA.

cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
```diff
@@ -975,4 +975,679 @@ var MetricsMigrations = []SchemaMigrationRecord{
...
+				},
+				Engine: Distributed{
+					Database:    "signoz_metrics",
+					Table:       "metric_reduction_rules",
+					ShardingKey: "cityHash64(metric_name)",
+				},
+			},
+			CreateMaterializedViewOperation{
+				Database:  "signoz_metrics",
+				ViewName:  "samples_v4_mv",
```
srikanthccv, line 1305: We want to write the non-reduce to regular tables. It could either be direct write or this mv. Chose mv because it simplifies the data flow and cost of this mv is minimal.

cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
```diff
@@ -975,4 +975,679 @@ var MetricsMigrations = []SchemaMigrationRecord{
...
+				// cumulative counters reduced to per-bucket increments with per-point
+				// reset detection (a value drop counts the post-reset value as the
+				// increment). the scan reaches one minute further back to supply the
+				// previous point at the window edge; a series' first point yields no
+				// increment (Prometheus increase() semantics). the output is labeled
+				// Delta temporality because the reduced per-bucket values are deltas
+				// (the reduced series catalog MV relabels these series to match).
+				Query: `SELECT
+							env,
+							'Delta' AS temporality,
```
srikanthccv, line 1580: See the https://app.notion.com/p/signoz/Cardinality-reduction-379fcc6bcd19807e8914f5daf434888c?source=copy_link#379fcc6bcd19818a9b67ef058520bae6

srikanthccv: However, the `__temporality__` label still maintains the orignal temporality.

exporter/signozclickhousemetrics/batch.go
```diff
@@ -10,19 +10,23 @@ import (
 )
 
 type batch struct {
-	samples  []*sample
-	expHist  []*exponentialHistogramSample
-	ts       []*ts
-	metadata []*metadata
-	logger   *zap.Logger
+	samples  []sample
```
srikanthccv, line 13: With `[]*sample`, the pointer `&sample{}` gets stored into the slice, which outlives the call in exporter. This means the `sample` spills to the heap and a heap object per element is created. With `[]T`, `*t` copies the struct into the backing array; the pointer is only dereferenced, never stored, so the literal doesn't escape and lives on the stack. All the elements end up in one backing array instead of N separate heap objects.

So the change is from N per-element heap objects to 1 array allocation.

How did it matter? The gc cost is proportional with the number of live objects it has to track, scan, and sweep. Reduce N tiny objects into one contiguous has shown reduced GC-pressure. It was a major contributor to GC dropping from ~39% to ~17% of CPU over the whole effort. The change was more visible in addMetadata because it add more elements. Overall reduced allocs and less gc pressure.


Storing values means the array grows by copying (memmove) as it doubles. On its own, the change traded allocation for allocated bytes, and indeed bytes actually increased at first because the metadata array grew-and-copied. Pairing it with the adaptive presize (size each slice from the previous batch's length, so the array is allocated once at the right size) removed that churn.

exporter/signozclickhousemetrics/config.go
```diff
@@ -24,6 +25,27 @@ type Config struct {
 	TimeSeriesTable string `mapstructure:"time_series_table"`
 	ExpHistTable    string `mapstructure:"exp_hist_table"`
 	MetadataTable   string `mapstructure:"metadata_table"`
+
+	Reduction ReductionConfig `mapstructure:"reduction"`
+
+	// MetadataWriteSampleRatio, in (0, 1], is the fraction of metadata rows
+	// written per batch; 1.0 (default) writes all. Opt-in; lowering it trades
+	// attribute-catalog completeness for fewer writes at extreme ingest.
+	MetadataWriteSampleRatio float64 `mapstructure:"metadata_write_sample_ratio"`
```
srikanthccv, line 34: To help with https://github.com/SigNoz/engineering-pod/issues/5059. 

The NW is specific in that it has 4000 metrics that could have been handlful if they didn't move the dynamic part to name of the metric.

This means same attributes like `container.id` repeat across all the metrics, which leads to extreme write amplication that overwhlems the metadata spool and eventually throwing error.

There are broadly three approaches.

- in batch dedup; incorporated
- across batch dedup; more memory and the first, last reported won't be accurate.
- random sample on high ingest system


The default is what we do today, take all meta entries and write them. Where it's needed to be tweaked, the sampling can be adjust that it's both useful in suggestions for quick filters and other query parts without overwhelming the sytem.

This could be more robust such per-key sampling or version which is hueristic based on distinct values count but started of the with this universal simple sampling as it shown good in testing.

exporter/signozclickhousemetrics/exporter.go
```diff
@@ -618,11 +724,11 @@ func (c *clickhouseMetricsExporter) processHistogram(b *batch, metric pmetric.Me
 	b.addMetadata(name+sumSuffix, desc, unit, pmetric.MetricTypeSum, temporality, isMonotonic, resourceFingerprint, firstSeenUnixMilli, lastSeenUnixMilli)
 	b.addMetadata(name+sumSuffix, desc, unit, pmetric.MetricTypeSum, temporality, isMonotonic, scopeFingerprint, firstSeenUnixMilli, lastSeenUnixMilli)
 
-	b.addMetadata(name+minSuffix, desc, unit, pmetric.MetricTypeGauge, pmetric.AggregationTemporalityUnspecified, isMonotonic, resourceFingerprint, firstSeenUnixMilli, lastSeenUnixMilli)
-	b.addMetadata(name+minSuffix, desc, unit, pmetric.MetricTypeGauge, pmetric.AggregationTemporalityUnspecified, isMonotonic, scopeFingerprint, firstSeenUnixMilli, lastSeenUnixMilli)
+	b.addMetadata(name+minSuffix, desc, unit, pmetric.MetricTypeGauge, pmetric.AggregationTemporalityUnspecified, false, resourceFingerprint, firstSeenUnixMilli, lastSeenUnixMilli)
```
srikanthccv, line 727: existing issue, fixed here.

exporter/signozclickhousemetrics/rules.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+	c.reductionRules.Store(&rules)
+	if c.reductionActiveRules != nil {
+		c.reductionActiveRules.Record(ctx, int64(len(rules)))
+	}
+}
+
+func (c *clickhouseMetricsExporter) fetchReductionRules(ctx context.Context) (ruleSet, error) {
+	// argMax over the distributed table picks the latest rule regardless of shard.
+	query := fmt.Sprintf(`SELECT
```
srikanthccv, line 85: only inserts into the rule table

srikanthccv: avoiding mutation

internal/common/fingerprint/label.go
```diff
@@ -1,65 +1,92 @@
...
+	"sync"
 )
 
+const nameLabel = "__name__"
+
+// Pooled scratch space for the key slice and output buffer.
+var (
+	labelKeysPool = sync.Pool{New: func() any { s := make([]string, 0, 64); return &s }}
+	labelBufPool  = sync.Pool{New: func() any { b := make([]byte, 0, 512); return &b }}
+)
```
srikanthccv, line 14: take inspiration from upstream `prometheusexporter` pool impl; sort of one-to-one

cmd/signozschemamigrator/schema_migrator/metrics_migrations.go
```diff
@@ -975,4 +975,679 @@ var MetricsMigrations = []SchemaMigrationRecord{
...
+										row_number() OVER rate_window = 1, nan,
+										value < lagInFrame(value, 1) OVER rate_window, value,
+										value - lagInFrame(value, 1) OVER rate_window
+									) AS increment
+								FROM signoz_metrics.samples_v4_buffer
+								PREWHERE reduced_fingerprint != 0
+								WHERE (bitAnd(flags, 1) = 0)
+									AND ((temporality = 'Cumulative') AND (is_monotonic = true))
+									AND (unix_milli < (toUnixTimestamp(now() - toIntervalSecond(120)) * 1000))
+									AND (unix_milli >= (toUnixTimestamp(now() - toIntervalMinute(6)) * 1000))
```
srikanthccv: discrepancy
```
								AND (unix_milli < (toUnixTimestamp(now() - toIntervalSecond(120)) * 1000))
								AND (unix_milli >= (toUnixTimestamp(now() - toIntervalMinute(11)) * 1000))

```

exporter/signozclickhousemetrics/batch.go
namanv3: Makes sense, saw something similar in formula calculation on the main repo as well

therealpandey (approved):

srikanthccv:
@namanv3, please continue the review and understand the change in depth when you get time.

PR #848, opened 2026-06-23
feat: support for traces simulation

nityanandagohain:
### Summary
Adds support for trace simulation, though which we can send spans and then simulate how the collector processors would transform it.

### Related Issues

Part of https://github.com/SigNoz/engineering-pod/issues/5267

### Resource Impact
- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collectorsimulator/collectorsimulator.go
+++ b/pkg/collectorsimulator/collectorsimulator.go
@@ -14,6 +14,7 @@ import (
 	"go.opentelemetry.io/collector/confmap/provider/envprovider"
 	"go.opentelemetry.io/collector/confmap/provider/fileprovider"
 	"go.opentelemetry.io/collector/otelcol"
+	"go.opentelemetry.io/collector/pipeline"
 	"go.opentelemetry.io/collector/processor"
 	"go.opentelemetry.io/collector/service"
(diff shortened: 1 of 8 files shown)


pkg/collectorsimulator/traces_test.go
```diff
@@ -0,0 +1,147 @@
+package collectorsimulator
```
therealpandey, line 1: @nityanandagohain please check the usages of require v/s assert.

pkg/collectorsimulator/collectorsimulator.go
```diff
@@ -40,8 +40,16 @@ type CollectorSimulator struct {
 
 type ConfigGenerator func(baseConfYaml []byte) ([]byte, error)
 
+type Signal string
```
therealpandey: Can we use the signal from otel? Upstream core otel has an enum defined already

therealpandey (changes requested):

pkg/collectorsimulator/traces_test.go
nityanandagohain: thanks have updated to assert in the final checks.

therealpandey (approved):

PR #850, opened 2026-06-24
feat(trace-json): reapply - store trace attribute as json

krsoninikhil:
### Summary
Store attributes column of json type to store the attributes as raw json. Needs clickhouse > v25.12.5

### Related Issues

Part of https://github.com/SigNoz/engineering-pod/issues/5068

### Resource Impact
- [ ] Yes
- [x] No


exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -171,6 +171,51 @@ func populateEventsV3(events ptrace.SpanEventSlice, span *SpanV3, lowCardinalExc
...
+			zap.String("event_span_id", utils.SpanIDToHexOrEmptyString(spanID)),
+		)
+		return "{}"
+	}
+	return string(b)
+}
+
+// sanitizeJSONFloats hanldes NaN/Inf float64 values, encoding/json errors out
+// for these values, so a single malformed double attribute could not fail the whole span.
+func sanitizeJSONFloats(v any) any {
```
nityanandagohain, line 197: we already go through the entire attribute in FlattenJSON inside the `add` function. Check if we can decrese cost of our processing by doing this there itself.

nityanandagohain (commented):
one comment regarding traversing the map

exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
krsoninikhil: This function is only called when marshaling actually [fails](https://github.com/SigNoz/signoz-otel-collector/pull/850/changes#diff-ae998f51c057cf20f412971664f275d56f437f4b9ef729c23a0fca2577e49da4R181), so no cost for normal cases. Kept it separate to avoid mixing with map related logic incase we deprecate that eventually, then this path won't be touched.

nityanandagohain (dismissed):
looks, good. Do monitor while releasing in a phased manner.

exporter/clickhousetracesexporter/clickhouse_exporter.go
nityanandagohain, line 111: there should be an extra `?` here ?

nityanandagohain (changes requested):

exporter/clickhousetracesexporter/clickhouse_exporter.go
krsoninikhil: Ah, thanks for catching this. Insert were still working fine without this 🤔 .

Fixed in [54b30b1](https://github.com/SigNoz/signoz-otel-collector/pull/850/commits/54b30b132409dbcc82a062065175976c7abe5885)

nityanandagohain: Interesting, I faced a similar issue before where it worked but forgot the reason.

nityanandagohain (approved):

PR #852, opened 2026-06-27
fix(exporter/signozclickhousemetrics): do not count reduced metrics towards usage and other fixes

srikanthccv:
Follow up to #839 

- exclude reduced metrics from usage count, and report the retained samples separately
- fix the issue with histogram `.bucket` reduced rule not working, the metric name is further normalised
- add the missing 1.5x we discussed, good it works without it but as an added benefit, update
- update the usage to report from the reduced tables in hourly fashion, with 2 windows look back to tolerate 2-3 hours of delay in ingesting data (under regular circumstances). it can always be rectified if needed.

### DIFF
--- a/usage/usage.go
+++ b/usage/usage.go
@@ -15,6 +15,20 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	// 10m refreshable mvs + 5 minutes grace
+	reducedUsageSettleWindow = 15 * time.Minute
+	reducedUsageTimeout      = 120 * time.Second
+	// report every hour window
+	reducedUsageWindow = time.Hour
+	// grace look back
+	reducedUsageLookbackWindows = 3
+)
+
+var reducedUsageNamespace = uuid.MustParse("6f9619ff-8b86-d011-b42d-00cf4fc964ff")
+
+var reducedUsageCollectorID = uuid.MustParse("00000000-0000-0000-0000-7265647563ed")
+
 // Options provides options for LogExporter
 type Options struct {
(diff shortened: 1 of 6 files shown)


usage/usage.go
```diff
@@ -15,10 +15,29 @@ import (
...
+	// 10m refreshable mvs + 5 minutes grace
+	reducedUsageSettleWindow = 15 * time.Minute
+	reducedUsageTimeout      = 120 * time.Second
+	// report every hour window
+	reducedUsageWindow = time.Hour
+	// grace look back
+	reducedUsageLookbackWindows = 3
+)
+
+var reducedUsageNamespace = uuid.MustParse("6f9619ff-8b86-d011-b42d-00cf4fc964ff")
```
vikrantgupta25, line 28: hardcoded ?

therealpandey (approved):

vikrantgupta25 (changes requested):

usage/usage.go
therealpandey: 🤣

srikanthccv: right, this is because the exporter ids needs to be deterministic across the pods. there will be n replicas of the collector pods each writing the same that that will deduped at the query time. let me know if this clears it up or if I should elaborate. 

if it were done outside the collector with a standalone process, then none of this fixed values is needed.

vikrantgupta25: got it makes sense

vikrantgupta25 (approved):

PR #866, opened 2026-07-28
refactor(schemamigrator): run body_v2 migrations unconditionally

tushar-signoz:
Removes the `ENABLE_LOGS_MIGRATIONS_V2` env gate so logs migration **2001** (`body_v2`, `body_promoted`, `signoz_metadata.field_keys`) runs in the normal flow.

### Why it's safe

- `LogsMigrationsV2` was a **strict superset** of `LogsMigrations` — 1000–1005 were identical, so the swap re-runs nothing and skips nothing on existing clusters.
- The ClickHouse features 2001 needs are **already ungated**: traces 1011–1013 set the same `object_serialization_version='v3'` / `'advanced'` settings and add a JSON column with a `JSONAllPaths` index; logs 1004 already adds a JSON column.
- All of 2001 is sync/lightweight (metadata-only column adds; indexes cover new data only), and writes stay behind `body_json_enabled` (default off), so the columns land empty. Near-zero cost on existing clusters.
- **MigrationID 2001 is unchanged**, so clusters that ran it with the flag on don't re-run it or orphan their `schema_migrations_v2` row.

### Also in here

- **Rollback fix:** 2001 dropped `body_v2`/`body_promoted` from `logs_v2` but not `distributed_logs_v2`, leaving the distributed table referencing missing columns. Now drops distributed first, then local.
- **Rename:** `logs_migrations_v2.go` → `logs_migrations.go` (and its test), since the `_v2` suffix referred to the removed flag. The other `v2` filenames are untouched — their `v2` refers to the `schema_migrations_v2` table, not this flag.

### ⚠️ Rollout ordering

`getLastSyncMigration` now returns 2001 instead of 1005, so the collector's `migrate sync check` gate waits for 2001 to be `Finished`. **The migrator must roll out before the collector image**, otherwise new collector pods crashloop until the migrator catches up.

### Verified

`go build ./...`, `go test ./cmd/...`, `go vet`, `gofmt` — all clean. Generated up/down SQL inspected directly, since `DownItems` have no test coverage today.


tushar-signoz:
## Code review

Reviewed the diff against `main` and traced the migration framework to check the claims in the description. **No blockers** — the safety argument holds and I was able to independently confirm each load-bearing part of it. Notes below are mostly documentation/follow-up, plus one test nit.

### Claims I verified

**Superset claim ✅** — `diff` of `main:logs_migrations.go` against `main:logs_migrations_v2.go` shows 1000–1005 are identical except `utils.ToPointer(uint(100))` → `utils.ToPointer[uint](100)` (same type, same value — an explicit type argument vs. an inferred one). So swapping the lists re-runs nothing and skips nothing.

**Flag is fully gone ✅** — grep for `EnableLogsMigrationsV2` / `ENABLE_LOGS_MIGRATIONS_V2` / `LogsMigrationsV2` across the repo is clean. Nothing else read the flag, so no other behavior silently changes.

**ClickHouse-capability claim ✅** — traces `1006` added a JSON `resource` column to `signoz_index_v3` and traces `1011` later flipped the *same two* settings (`object_serialization_version='v3'`, `object_shared_data_serialization_version='advanced'`) on that same table, ungated and already shipped. Logs `1004` (JSON `resource`) + `2001` (settings flip) is the identical sequence, so nothing new is being required of the server.

**Async check is unaffected ✅** — worth stating explicitly since only the sync gate is called out in the description. `MigrationManager.IsAsync` returns true only if *some* `UpItem` is async, and all twelve of 2001's up-items are sync (`CreateTableOperation` / `AlterTableAddColumn` / `AlterTableAddIndex` / `AlterTableModifySettings` are all non-mutation + idempotent + lightweight). So `getLastAsyncMigration(LogsMigrations)` still returns 1003 — only `getLastSyncMigration` moves 1005 → 2001, exactly as described.

**`signoz_metadata` exists before 2001 runs ✅** — `Bootstrap()` → `createDBs()` (`manager.go:149`) issues `CREATE DATABASE IF NOT EXISTS` for every entry in `Databases`, including `signoz_metadata`, before any migration list is walked. So 2001 creating `field_keys` in the metadata DB from inside the *logs* list is safe on a fresh cluster.

**Rollback fix is correct ✅** and lands in the path that actually executes it. All of 2001's `DownItems` are mutations (`AlterTableDropIndex`, `AlterTableDropColumn`, `DropTableOperation` all return `IsMutation() == true`), so `MigrateDownSync` skips 2001 entirely and the whole rollback runs through `MigrateDownAsync`, which iterates `DownItems` in declaration order sequentially — the distributed-before-local ordering holds. It also brings 2001 in line with traces `1012`, whose `DownItems` already drop `distributed_signoz_index_v3` before `signoz_index_v3`.

### Notes

**1. The settings flip isn't purely "new columns land empty."** `AlterTableModifySettings` (`logs_migrations.go:310`) is table-level, so after 2001 new `logs_v2` parts serialize the *pre-existing* `resource` JSON column (from 1004) with v3/advanced too — independent of `body_json_enabled`. This is the one part of 2001 whose effect isn't gated by the write flag. I think it's genuinely fine (existing parts keep their format, reads are version-aware, and traces 1006→1011 is exact shipped precedent), but the "columns land empty → near-zero cost" framing in the description reads as if nothing touches existing data, and an on-call reading it later would be misled. Worth one clarifying line.

**2. Rollback leaves the settings in place.** `DownItems` doesn't revert `object_serialization_version` / `object_shared_data_serialization_version`. I believe that's the right call — you can't un-serialize parts already written in v3 — and traces 1011 has empty `DownItems` for the same reason. But since this PR is specifically fixing the down path, a one-line comment saying the settings flip is intentionally not reverted would stop the next reader from filing it as the same class of bug you just fixed.

**3. No kill switch remains.** The flag was the only way to hold 2001 back on a specific cluster. Afterwards the only lever is `--up` version pinning on the schema-migrator, and the collector's own `migrate sync up` path doesn't thread `upVersions` through at all — so on a collector-driven rollout there's no way to skip 2001. That's the intended outcome and the cost argument supports it; just worth stating in the description so nobody goes looking for the flag during an incident. Related: the helm chart / SaaS config may still set `ENABLE_LOGS_MIGRATIONS_V2` — that becomes inert rather than broken, but it's worth a cleanup issue in the chart repo.

**4. Test nit — reference records by ID, not slice index.** `LogsMigrations[6], // 2001 (sync)` silently points at the wrong record if anyone inserts a migration mid-list, and `checkSchemaMigrationRecordExactNature` only asserts that the records handed to it are sync/async — it never checks coverage — so a shifted index keeps passing while testing something else. The old `TestLogsMigrationsV2ExactNature` is evidence this happens: it never asserted 2001's nature at all. The new test closes that gap, but a by-ID lookup would make the gap structurally impossible:

```go
func migrationByID(t *testing.T, migrations []SchemaMigrationRecord, id uint64) SchemaMigrationRecord {
	t.Helper()
	for _, m := range migrations {
		if m.MigrationID == id {
			return m
		}
	}
	t.Fatalf("migration %d not found", id)
	return SchemaMigrationRecord{}
}
```

**5. Rollout ordering** — agreed and well flagged. One refinement: the risk is specifically a *separate* schema-migrator image/job lagging behind the collector. If `migrate sync up` runs from the same collector image (init container), the two are self-consistent and the ordering takes care of itself.

### Unrelated, spotted nearby

`runSquashedMigrations` (`cmd/signozotelcollector/migrate/sync_up.go:164`) does `return nil` instead of `continue` when `ShouldRunSquashedV2` is false, so the first DB that doesn't need squashed migrations aborts the loop and the other two are skipped — and since it iterates a map, which DB that is varies per run. Pre-existing and out of scope here; flagging so it doesn't get lost.

### Local verification

`go build ./...` clean; `go test ./cmd/signozschemamigrator/...` passes.

therealpandey (approved):

PR #883, opened 2026-08-17
feat(signozlogspipeline): infer otlp fields the log carries in its body

tushar-signoz:
### Summary

`normalize` shapes a log body into a map with a canonical `message` field. It can now also fill in the top level otlp fields that a log carries *inside* itself. The problem it solves is SigNoz/signoz#5245: users ship JSON logs with the level, the trace id and the logger name in the body, rarely add the pipeline steps that would lift them out, and end up with an empty `severity_text`, logs that aren't correlated to their traces, and no scope.

A target the record does not already carry is looked for in four containers, in order — the JSON body, the log attributes, the instrumentation scope attributes, the resource attributes — under the field names configured for it. Names match case insensitively, and when two keys in one container differ only in case the lexicographically smaller one wins so the result never depends on map order.

**The collector's own defaults leave every list except `message_fields` empty, and an empty list turns that target's inference off.** A bare `type: normalize` therefore behaves exactly as it does today — message promotion only — and inference activates when SigNoz sends the field lists over opamp. The lists query-service is expected to send:

| target | setting | field names SigNoz sends | collector default |
| --- | --- | --- | --- |
| `message` in the body | `message_fields` | — | `message`, `body`, `log`, `msg` |
| severity_number | `severity_number_fields` | `severity_number`, `severitynumber` | empty (off) |
| severity_text | `severity_text_fields` | `severity_text`, `severitytext`, `severity`, `level`, `log.level`, `log_level`, `loglevel`, `levelname`, `lvl` | empty (off) |
| trace_id | `trace_id_fields` | `trace_id`, `traceid`, `trace.id` | empty (off) |
| span_id | `span_id_fields` | `span_id`, `spanid`, `span.id` | empty (off) |
| trace_flags | `trace_flags_fields` | `trace_flags`, `traceflags`, `trace.flags` | empty (off) |
| scope name | `scope_name_fields` | `scope.name`, `scope_name`, `scopename` | empty (off) |
| scope version | `scope_version_fields` | `scope.version`, `scope_version`, `scopeversion` | empty (off) |

The lists are the only switch — there are no separate enable/disable flags, and the first name in a list that a log carries wins. The two halves of the severity are read independently, each from its own fields and each only when the record does not already carry that half; whichever half is still missing at the end is filled in from the other, so a `level` alone names the number and a `severity_number` alone names the text.

A level name the operator knows records the level *group*, so `warning`, `Warning` and `WARN` all become `WARN`, and syslog's `critical` becomes `ERROR` with severity_number 18. A name it doesn't know is recorded as the log wrote it and names no severity_number, so an app's own `AUDIT` level survives instead of the log arriving with no severity at all. That means a field holding prose rather than a level name is recorded in full into a LowCardinality column; capping it is tracked separately.

### Mental model: one shopping trip per log

Think of inference as a single shopping trip the operator makes for each log record:

1. **The shopping list is written per record** — `infer()` (fields.go) builds a `wantedFields` array, one slot per target (`targetSeverityText`, `targetTraceID`, …). A slot is filled only if the record doesn't already carry that field (`ent.SeverityText == ""`, `len(ent.TraceID) == 0`, …) — that's the never-overwrite guarantee, decided before anything is searched. What goes in the slot is that target's `parseFunc` (see 5).

2. **Stores are visited in a fixed order** — `searchOrder()` returns the four containers as `[4]map[string]any`: JSON body → `ent.Attributes` → scope attributes → `ent.Resource`. The store closest to the log wins.

3. **One walk per store, not one per item** — `scan()` walks each container's keys once and checks every key against the whole list simultaneously. This works because all configured names live in one lookup table built once at config time: `newFieldConfig()` folds every `*_fields` list into `fieldConfig.names`, a `map[string][]nameMatch` from lowercased key name → which targets claim it and at what preference.

4. **Each item has ranked brand preferences** — `nameMatch.rank` is the position in the `*_fields` list, so `severity_text_fields: [level, lvl]` means `level` (rank 0) beats `lvl` (rank 1) when one store has both. Inside `scan()`, `bestRank` and `bestKey` track the winner per target; a tie between same-rank keys goes to the lexicographically smaller key, so the outcome never depends on Go map iteration order.

5. **Quality check happens at the shelf** — the target's `parseFunc` must accept the value: `parseTraceID`/`parseSpanID` demand correctly-sized non-zero hex, `parseTraceFlags` a byte, `parseSeverityNumber` something in 1–24, `parseNonEmptyString` a non-blank string. A value that fails is passed over and the hunt for that item continues, in this store and the next.

6. **An item found is crossed off** — `scan()` nils that target's `wanted` slot after each container and decrements `remaining`; later stores are never searched for it, and when `remaining` hits 0 the trip ends early.

7. **Checkout: copy or take** — the setters call `take(results, target)`, which hands over the found value and consults `fieldConfig.move[target]` (fed by `move_all_fields` / `move_*_field`): `false` records the value and leaves the original on the shelf (copy, the default); `true` calls `result.source.remove()`, deleting the key from whichever container it came from.

8. **Severity is two half-items that complete each other** — `setSeverity()` (severity.go) records the number and the text independently, then reconciles: a missing number is derived from the text via `severityByLevelName` (so `level: warning` alone names 13), and a missing text is derived from the number via `severityText()` (so a bare severity_number 9 also names `INFO`). The reconciliation runs even when the trip bought nothing, which is why a producer-set number gets its text filled in. `setTraceContext()` and `setScope()` are plain checkouts with no such step.

9. **An empty field list means the item is never on the list** — `addNames()` registers nothing for it, so no key can ever match: inference off for that target. That's the collector default (`NewConfigWithID`) for everything except `message_fields`, so a bare `type: normalize` shops only for the message until SigNoz sends the lists over opamp.

(`message` itself is the one item handled before the trip: `promoteMessage()` (transformer.go) runs a solo `scan()` for `targetMessage` with `anyValue` as the parse — any non-nil value qualifies — and always removes the source key, never copies.)

### Copy first, move later

A value the record already carries is never overwritten, so an explicit `severity_parser` or `trace_parser`, or anything the producer set, always wins and nothing is removed in that case.

A field that is used is **copied**: the key stays where it was found. This is deliberate for the first rollout — `body.level` and `attributes.trace_id` keep working for saved views, dashboards and alerts that filter on them. `message` is the one exception, and is not configurable: promoting a field into `body.message` has always removed it.

`move_all_fields` removes the key that was used instead, and takes precedence over the per field `move_severity_number_field`, `move_severity_text_field`, `move_trace_id_field`, `move_span_id_field`, `move_trace_flags_field`, `move_scope_name_field` and `move_scope_version_field`, which all default to `false`. It exists so that a pipeline wanting everything moved doesn't have to enumerate the targets that exist today: a target added later is covered automatically, instead of being copied while the listed ones are moved.

Moving is where this is headed, once we've checked that nobody depends on the body paths it removes.

### Activation over opamp

Query-service currently emits a bare `type: normalize` operator, so nothing changes at merge time; the SigNoz-side change to send the field lists (a `pipelinetypes.PipelineOperator` addition) is what turns inference on. Two things verified about that path:

- The collector applies remote config **by content**, not by pipeline version: `opamp/config_manager.go` hashes the received YAML bytes and reloads whenever they differ, regardless of the server-side `ConfigHash` (which encodes the pipeline version). Sending the lists takes effect without a new pipeline version.
- Empty-by-default makes the rollout order safe: an agent that never receives the lists just keeps today's behaviour. The only skew to gate on is genuinely old collector builds that don't know these keys at all — their strict confmap rejects configs containing them.

### Fields the processor used to lose

`entry.Entry`, the stanza type every operator works on, is a subset of an otlp log record: it has a scope *name* and nothing else of the scope, and no event name. Anything it can't hold was silently dropped by this processor on every log that passed through it. Scope version, scope attributes and event name are now carried across the conversion in internal temp attributes and restored on the way out, and `ScopeLogs` grouping keys on the whole scope identity rather than the name alone.

Still dropped, for anyone who needs them later: `resource.schema_url` and `scope.schema_url` (which `schemaprocessor`, part of this collector, keys off), and `dropped_attributes_count`.

### Related Issues

Ref SigNoz/signoz#5245

### Resource Impact
- [x] Yes
- [ ] No

If **Yes**, describe:
- **Resources impacted:** CPU, and only once the field lists are sent — with the default (empty) config the added cost is negligible. No change to memory beyond a few short lived allocations per record, and none to disk or network.
- **Expected change:** about **+1.2us of CPU per log record** with the full lists configured. Benchmarked on an M4 Pro, against a log with 8 body keys and 21 resource attributes:

  | scenario | ns/op | allocs/op |
  | --- | --- | --- |
  | inference off (the collector default) | 1,220 | 5 |
  | typical json log: `msg` + `level` in the body | 2,364 | 6 |
  | every target present in the body | 2,415 | 14 |
  | no target present anywhere | 2,400 | 2 |

  At 100k logs/s that is roughly 0.12 of a core.

- **Benchmarks:** the numbers above were measured with a benchmark harness kept out of this PR to keep it lean.

  The first cut of this cost ~7.0us on the typical log, because each target walked all four containers until it found a value and most targets are absent on a real log, so the cost was dominated by re-walking the resource attributes once per target. Looking for every target in a single pass over each container brought it to the numbers above: 725ns for the walk when every target matches, 463ns and zero allocations when none do.

### Verified end to end

Against the devenv ClickHouse with a collector built from this branch (field lists configured), 15 log shapes were ingested and read back out of `logs_v2`: level/severity_number in the body, trace id, span id and trace flags (hex and numeric), scope name and version, message promotion from `msg`/`log`/attributes, a nested map body, syslog `critical`, an unknown level, an all zero trace id, and a producer set `severityText`. The same battery was replayed after the single pass rewrite and every row was byte identical, except the unknown level case that this PR deliberately changed.


tushar-signoz (commented):
Function-by-function notes on what each new piece does.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	target fieldTarget
+	rank   int
+}
+
+type fieldConfig struct {
+	names map[string][]nameMatch
+	move  [targetCount]bool
+}
+
+func newFieldConfig(c Config) fieldConfig {
```
tushar-signoz, line 42: Builds the one-time lookup table at operator start: every configured field name, lowercased, mapped to the target(s) that claim it and their preference order. Also freezes each target's copy-vs-move decision. Everything per-record reads this, nothing mutates it.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+		targetSpanID:         c.MoveAllFields || c.MoveSpanIDField,
+		targetTraceFlags:     c.MoveAllFields || c.MoveTraceFlagsField,
+		targetScopeName:      c.MoveAllFields || c.MoveScopeNameField,
+		targetScopeVersion:   c.MoveAllFields || c.MoveScopeVersionField,
+	}
+
+	return f
+}
+
+func (f *fieldConfig) addNames(target fieldTarget, names []string) {
```
tushar-signoz, line 67: Registers one target's name list into the shared table. Position in the list becomes the name's rank (earlier = preferred); blanks and duplicate names for the same target are skipped.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+	return false
+}
+
+func normalizeFieldName(name string) string {
+	return strings.ToLower(strings.TrimSpace(name))
+}
+
+func searchOrder(ent *entry.Entry) [4]map[string]any {
```
tushar-signoz, line 92: The four places a field can live, in priority order: JSON body, log attributes, scope attributes, resource attributes. scan() walks them left to right, so the container closest to the log wins.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	value  any
+	source fieldSource
+	found  bool
+}
+
+type scanResults [targetCount]scanResult
+
+// take applies the move setting for a target that was found, and reports the value so the
+// caller can record it.
+func (f fieldConfig) take(results scanResults, target fieldTarget) (any, bool) {
```
tushar-signoz, line 113: Post-scan checkout for one target: hands back the winning value if the scan found one, and deletes the source key when this target is configured to move rather than copy.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	return result.value, true
+}
+
+// scan walks each container once, looking every key up against the names of all wanted
+// targets at once, rather than walking the containers again for each target. A container is
+// searched only while some target is still unresolved, and a target is resolved by the first
+// container that yields a usable value for it. Within a container the lowest ranked name
+// wins, ties going to the lexicographically smaller key so the result never depends on map
+// order, and a value the target can't use is passed over in favour of the next candidate.
+func (f fieldConfig) scan(containers [4]map[string]any, wanted wantedFields) scanResults {
```
tushar-signoz, line 130: The single pass. Walks each container's keys once, matching every key against all still-unresolved targets at the same time instead of re-walking the containers per target. A target is resolved by the first container that yields a value its parse function accepts; within a container the lowest-ranked name wins and rank ties go to the lexicographically smaller key so map iteration order never shows. Stops as soon as nothing is left to find.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+
+	return results
+}
+
+// infer fills in the top level fields the log carries inside itself. Every target is looked
+// for in the same pass, so a record costs one walk of its containers rather than one per
+// target, and what each container holds is read before any of it is moved out.
+func (f fieldConfig) infer(ent *entry.Entry) {
```
tushar-signoz, line 186: Per-record entry point for inference: lists which targets this record is still missing (never-overwrite lives here), runs one scan for all of them, then hands the results to the severity/trace/scope setters.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
```diff
@@ -0,0 +1,127 @@
+package json
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/entry"
+)
+
+var severityByLevelName = func() map[string]entry.Severity {
```
tushar-signoz, line 10: The level-name dictionary: every spelling we recognize (warn, warning, severe, crit, emerg, ...) mapped to its otlp severity number. Built once at package init from the per-severity name lists.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
```diff
@@ -0,0 +1,127 @@
...
+	severities := map[string]entry.Severity{}
+	for severity, names := range levelNames {
+		for _, name := range names {
+			severities[name] = severity
+		}
+	}
+	return severities
+}()
+
+func (f fieldConfig) setSeverity(ent *entry.Entry, results scanResults) {
```
tushar-signoz, line 47: Records the two severity halves independently, then reconciles: whichever half is still missing at the end is derived from the other, so a level name alone also yields the number and a bare number also yields a text. A text the dictionary doesn't know is kept as the log wrote it and names no number. The reconciliation (last two ifs) runs even when the scan found nothing, which is how a producer-set severity_number gets its text filled in.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
```diff
@@ -0,0 +1,127 @@
...
+		if severity, ok := severityFromLevelName(ent.SeverityText); ok {
+			ent.Severity = severity
+		}
+	}
+	if ent.SeverityText == "" && ent.Severity != entry.Default {
+		ent.SeverityText = severityText(ent.Severity)
+	}
+}
+
+func severityText(severity entry.Severity) string {
```
tushar-signoz, line 74: Names a severity number's group: 9 -> INFO, 13 -> WARN. The four levels inside each group all take the group's name, matching how known level names are recorded.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
```diff
@@ -0,0 +1,127 @@
...
+
+func parseSeverityNumber(value any) (any, bool) {
+	severity, ok := severityFromNumber(value)
+	if !ok {
+		return nil, false
+	}
+	return severity, true
+}
+
+func severityFromNumber(value any) (entry.Severity, bool) {
```
tushar-signoz, line 99: Accepts a severity number in the shapes logs actually carry: int, whole float (JSON numbers), or string - a numeric string, or a level name / SEVERITY_NUMBER_-prefixed form. Anything outside 1-24 is rejected so garbage never lands in the column.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/trace.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/entry"
+)
+
+const (
+	traceIDSize    = 16
+	spanIDSize     = 8
+	traceFlagsSize = 1
+)
+
+func (f fieldConfig) setTraceContext(ent *entry.Entry, results scanResults) {
```
tushar-signoz, line 15: Plain checkout: writes whichever of trace_id, span_id and trace_flags the scan found. No reconciliation step here, unlike severity.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/trace.go
```diff
@@ -0,0 +1,100 @@
...
+
+func parseTraceID(value any) (any, bool) {
+	return parseID(value, traceIDSize)
+}
+
+func parseSpanID(value any) (any, bool) {
+	return parseID(value, spanIDSize)
+}
+
+func parseID(value any, size int) (any, bool) {
```
tushar-signoz, line 35: Shared validator for trace and span ids: exact-length hex string or raw bytes, and the all-zero id is rejected because otlp defines it as absent - recording it would make the log look correlated when it isn't.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/trace.go
```diff
@@ -0,0 +1,100 @@
...
+
+	for _, b := range id {
+		if b != 0 {
+			return id, true
+		}
+	}
+	return nil, false
+}
+
+func parseTraceFlags(value any) (any, bool) {
```
tushar-signoz, line 65: Trace flags arrive in the wild as hex ("01"), raw bytes, or a plain number; all collapse to the single W3C flags byte, and anything outside 0-255 is rejected.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/scope.go
```diff
@@ -0,0 +1,16 @@
+package json
+
+import (
+	signozstanzaentry "github.com/SigNoz/signoz-otel-collector/processor/signozlogspipelineprocessor/stanza/entry"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/entry"
+)
+
+func (f fieldConfig) setScope(ent *entry.Entry, results scanResults) {
```
tushar-signoz, line 10: Checkout for the scope pair. The name has a real field on entry.Entry; the version does not, so it rides in an internal temp attribute until utils.go restores it onto the emitted ScopeLogs.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/transformer.go
```diff
@@ -140,3 +127,18 @@ func (p *Processor) normalize(entry *entry.Entry) {
 		}
 	}
 }
+
+func (p *Processor) promoteMessage(ent *entry.Entry, message signozstanzaentry.Field) {
```
tushar-signoz, line 131: Message is looked up on its own, before inference: a solo scan for the first message-compatible field, any non-nil value accepted, and always moved - promoting a field into body.message has always removed the source key.

processor/signozlogspipelineprocessor/utils.go
```diff
@@ -12,15 +12,69 @@ import (
 	"go.opentelemetry.io/collector/pdata/plog"
 )
 
+type scopeID struct {
+	name       string
+	version    string
+	attributes uint64
+}
+
+func scopeIDOf(ent *entry.Entry) scopeID {
```
tushar-signoz, line 21: Identity of a scope for regrouping entries back into ScopeLogs: name + version + a hash of the scope attributes. Previously grouping keyed on the name alone, so two scopes differing only in version or attributes collapsed into one.

processor/signozlogspipelineprocessor/utils.go
```diff
@@ -12,15 +12,69 @@ import (
...
+	version, _ := ent.Attributes[signozstanzaentry.InternalTempScopeVersionAttribute].(string)
+	attributes, _ := ent.Attributes[signozstanzaentry.InternalTempScopeAttributesAttribute].(map[string]any)
+	return scopeID{
+		name:       ent.ScopeName,
+		version:    version,
+		attributes: adapter.HashResource(attributes),
+	}
+}
+
+func setScope(dest pcommon.InstrumentationScope, ent *entry.Entry) {
```
tushar-signoz, line 31: The way out: rebuilds the full scope on the emitted ScopeLogs - name from the entry, version and attributes from the temp attributes they were stashed in. upsertToMap skips the internal temp prefix, so the stash never leaks into record attributes.

processor/signozlogspipelineprocessor/utils.go
```diff
@@ -12,15 +12,69 @@ import (
...
+	dest.SetName(ent.ScopeName)
+	if version, ok := ent.Attributes[signozstanzaentry.InternalTempScopeVersionAttribute].(string); ok {
+		dest.SetVersion(version)
+	}
+	if attributes, ok := ent.Attributes[signozstanzaentry.InternalTempScopeAttributesAttribute].(map[string]any); ok {
+		upsertToMap(attributes, dest.Attributes())
+	}
+}
+
+func stashUnmappedFields(ld plog.Logs) {
```
tushar-signoz, line 41: The way in: entry.Entry cannot hold a scope version, scope attributes or an event name, and this processor used to silently drop all three. They are copied into internal temp attributes on each record before conversion, carried through the operators untouched, and restored on the way out.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+// as an any and asserted back by whoever asked for that target.
+type parseFunc func(value any) (any, bool)
+
+// wantedFields holds a parseFunc for each target still worth looking for on this record.
+type wantedFields [targetCount]parseFunc
+
+type nameMatch struct {
+	target fieldTarget
+	rank   int
+}
```
krsoninikhil: Lets add an inline comment for rank field for what is it.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (f *fieldConfig) addNames(target fieldTarget, names []string) {
+	rank := 0
+	for _, name := range names {
+		field := normalizeFieldName(name)
+		if field == "" || f.knows(target, field) {
+			continue
+		}
+		f.names[field] = append(f.names[field], nameMatch{target: target, rank: rank})
```
krsoninikhil: `knows` returns false if same field name exists with different target and it will override the map. Let's add that as a comment or via test if this is expected.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	return false
+}
+
+func normalizeFieldName(name string) string {
+	return strings.ToLower(strings.TrimSpace(name))
+}
+
+func searchOrder(ent *entry.Entry) [4]map[string]any {
+	body, _ := ent.Body.(map[string]any)
```
krsoninikhil, line 93: Does error case mean string body here? If string body should not reach here, then let's add a error log because body will be empty map on error.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	return [4]map[string]any{body, ent.Attributes, scopeAttributes(ent), ent.Resource}
+}
+
+func scopeAttributes(ent *entry.Entry) map[string]any {
+	attributes, _ := ent.Attributes[signozstanzaentry.InternalTempScopeAttributesAttribute].(map[string]any)
```
krsoninikhil: Does the error case denotes absence of scope attributes in the log? Then let's first check the presence of key in the map before type assertion and log the type assertion error.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/scope.go
```diff
@@ -0,0 +1,16 @@
...
+	if name, ok := f.take(results, targetScopeName); ok {
+		ent.ScopeName = name.(string)
+	}
+
+	if version, ok := f.take(results, targetScopeVersion); ok {
+		ent.AddAttribute(signozstanzaentry.InternalTempScopeVersionAttribute, version.(string))
```
krsoninikhil: Setting the version in the internal attribute? Is it not supported directly in the Entry?

processor/signozlogspipelineprocessor/utils.go
krsoninikhil: So we were not saving these scope name and version currently in logs? We have scope fields in the logs schema, are they not getting populated currently?

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
```diff
@@ -0,0 +1,127 @@
...
+		ent.Severity = severity.(entry.Severity)
+	}
+
+	if text, ok := f.take(results, targetSeverityText); ok {
+		name := text.(string)
+		if severity, known := severityByLevelName[strings.ToLower(name)]; known {
+			ent.SeverityText = severityText(severity)
+			if ent.Severity == entry.Default {
+				ent.Severity = severity
+			}
```
krsoninikhil, line 58: This part is already covered in next if block below?

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
```diff
@@ -0,0 +1,127 @@
...
+	if ent.SeverityText == "" && ent.Severity != entry.Default {
+		ent.SeverityText = severityText(ent.Severity)
+	}
+}
+
+func severityText(severity entry.Severity) string {
+	if severity == entry.Default {
+		return ""
+	}
+	group := entry.Severity((int(severity)-1)/4*4 + 1)
```
krsoninikhil, line 78: Hard to make changes later on relying on indexes. Lets rethink this if can be simplified with maps or string manipulations.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
```diff
@@ -0,0 +1,127 @@
...
+
+	switch v := value.(type) {
+	case int64:
+		number = v
+	case int:
+		number = int64(v)
+	case float64:
+		if v < float64(entry.Trace) || v > float64(entry.Fatal4) || v != float64(int64(v)) {
+			return entry.Default, false
+		}
```
krsoninikhil, line 110: These checks are added after switch as well, we can remove from here and move the additional check in that itself.

processor/signozlogspipelineprocessor/utils.go
```diff
@@ -12,15 +12,69 @@ import (
 	"go.opentelemetry.io/collector/pdata/plog"
 )
 
+type scopeID struct {
+	name       string
+	version    string
+	attributes uint64
```
krsoninikhil: Why ID suffix here?

And uint64 for attributes? If it's a presence check, then may be `hasAttributes`?

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	return attributes
+}
```
krsoninikhil: Can drop this single line method as well if only usage is in `searchOrder`.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	results := f.scan(searchOrder(ent), wanted)
+
+	f.setSeverity(ent, results)
+	f.setTraceContext(ent, results)
+	f.setScope(ent, results)
+}
+
+func anyValue(value any) (any, bool) {
+	return value, value != nil
+}
```
krsoninikhil: Lets drop this indirection.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	str, ok := value.(string)
+	if !ok {
+		return nil, false
+	}
+	str = strings.TrimSpace(str)
+	if str == "" {
+		return nil, false
+	}
+	return str, true
+}
```
krsoninikhil: All other parsers are defined in their own files except this.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+
+func scopeAttributes(ent *entry.Entry) map[string]any {
+	attributes, _ := ent.Attributes[signozstanzaentry.InternalTempScopeAttributesAttribute].(map[string]any)
+	return attributes
+}
+
+type fieldSource struct {
+	container map[string]any
+	key       string
+}
```
krsoninikhil, line 103: We can merge the `fieldSource` in `scanResult` itself.

processor/signozlogspipelineprocessor/processor.go
```diff
@@ -160,6 +160,7 @@ func (p *logsPipelineProcessor) startFromConverter() {
 
 func (p *logsPipelineProcessor) ConsumeLogs(ctx context.Context, ld plog.Logs) error {
 	p.telemetryBuilder.ProcessorIncomingItems.Add(ctx, int64(ld.LogRecordCount()), p.otelAttrs)
+	stashUnmappedFields(ld)
```
krsoninikhil, line 163: Question: Is it possible to do all the stashing in same place? [body stashing](https://github.com/SigNoz/signoz-otel-collector/pull/891) is happening in the transformer but scope fields will already be lost by that point 🤔

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/transformer.go
```diff
@@ -140,3 +127,18 @@ func (p *Processor) normalize(entry *entry.Entry) {
...
+	results := p.fields.scan(searchOrder(ent), wanted)
+	if !results[targetMessage].found {
+		return
+	}
+	if err := ent.Set(message, results[targetMessage].value); err != nil {
+		p.Logger().Error("Failed to set message field", zap.Error(err))
+		return
+	}
+	results[targetMessage].source.remove()
+}
```
krsoninikhil, line 144: Is this required separately now? Feels like a part of `infer` itself.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+// wantedFields holds a parseFunc for each target still worth looking for on this record.
+type wantedFields [targetCount]parseFunc
+
+type nameMatch struct {
+	target fieldTarget
+	rank   int
+}
+
+type fieldConfig struct {
+	names map[string][]nameMatch
```
krsoninikhil, line 38: Can there be multiple targets using same field for their value?

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+			remaining++
+		}
+	}
+
+	for _, container := range containers {
+		if remaining == 0 {
+			break
+		}
+
+		for key, value := range container {
```
krsoninikhil, line 148: This is iterating over each attribute of the container, that is not required if we iterate over the target and then over the potential fields to look for. Will probably also avoid using bestRank and bestKey maps. 

Might have to rethink the `fieldConfig.names` structure to map the targets to possible fields instead.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+			for _, match := range f.names[strings.ToLower(key)] {
```
krsoninikhil:
```suggestion
			for _, match := range f.names[normalizeFieldName(key)] {
```

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	if _, exists := ent.Attributes[signozstanzaentry.InternalTempScopeVersionAttribute]; !exists {
+		wanted[targetScopeVersion] = parseNonEmptyString
+	}
+
+	results := f.scan(searchOrder(ent), wanted)
+
+	f.setSeverity(ent, results)
+	f.setTraceContext(ent, results)
+	f.setScope(ent, results)
```
krsoninikhil, line 215: config object setting the Entry properties seem overloaded 🤔 

Let's keep the scope of config to hold the provided configuration and have another type like `NormalizedEntry` with fieldConfig as a property. Then all the set methods can belong to it.

krsoninikhil (changes requested):

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
krsoninikhil: Ah, if normalization of attributes is also required then we'll have to iterate over it anyway.

processor/signozlogspipelineprocessor/utils.go
krsoninikhil: Ah, its hash, so attributesHash?

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
tushar-signoz: Transform operator ensure body is map. That's why ignored error here.

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
tushar-signoz: This is standard following of severity_number. didn't invent it

processor/signozlogspipelineprocessor/utils.go
tushar-signoz: Ran the collector to ingest scope.version. it doesn't

processor/signozlogspipelineprocessor/processor.go
tushar-signoz: exactly

processor/signozlogspipelineprocessor/utils.go
tushar-signoz: done

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/severity.go
tushar-signoz: Slight difference on what value they are working on, severity vs severityText

krsoninikhil: This https://github.com/SigNoz/signoz-otel-collector/pull/883/changes/BASE..2cc509b2e8e097b805ae7b10c3c61e5f365650ef#diff-d172f3315c8ef922b518f32a5f34e00e4146f15537e10280f23f55ac92064d61R64-R68

processor/signozlogspipelineprocessor/utils.go
krsoninikhil: cc/ @nityanandagohain

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
tushar-signoz: `severity` field can be claimed by both `severity_text_fields` and `severity_number_fields`

processor/signozlogspipelineprocessor/stanza/operator/operators/normalize/fields.go
tushar-signoz: The naming can be better, config object is not really a config object . I'll think of a better name

krsoninikhil: Hm, it reads the config as well, so 🤷‍♂️ . If this is doing too much i.e. handling config and fixing the entry, breaking it can make testing easy and independent.

krsoninikhil (dismissed):
Please make the PR description consumable, it's too much for any human.

krsoninikhil (approved):

PR #888, opened 2026-08-18
test: fix flaky usage-stats race in logs and traces exporter tests

### LINKED ISSUE #634: (task) `TestExporterPushLogsData` Flaky behaviour
```bash
go test -timeout 30s -v  -run 'TestExporterInit|TestExporterPushLogsData' github.com/SigNoz/signoz-otel-collector/exporter/clickhouselogsexporter -count=1
```

TestExporterPushLogsData behaves flaky when testing the whole file together.
```bash
go test -timeout 30s -v  -run 'TestExporterInit|TestExporterPushLogsData' github.com/SigNoz/signoz-otel-collector/exporter/clickhouselogsexporter -count=1
=== RUN   TestExporterInit
at flush
at export metrics
at UsageExporter
--- PASS: TestExporterInit (0.10s)
=== RUN   TestExporterPushLogsData
at flush
at export metrics
at UsageExporter
    exporter_test.go:147: 
                Error Trace:    /Users/piyushsingariya/code/signoz-otel-collector/exporter/clickhouselogsexporter/exporter_test.go:147
                Error:          Expected nil, but got: &errors.errorString{s:"call to database Close, was not expected, next expectation is: ExpectedExec => expecting Exec or ExecContext which:\n  - matches sql: '.*insert into signoz_logs.distributed_usage.*'\n  - is without arguments"}
                Test:           TestExporterPushLogsData
    exporter_test.go:24: 
                Error Trace:    /Users/piyushsingariya/code/signoz-otel-collector/exporter/clickhouselogsexporter/exporter_test.go:24
                                                        /Users/piyushsingariya/code/signoz-otel-collector/exporter/clickhouselogsexporter/exporter_test.go:149
                Error:          Condition never satisfied
                Test:           TestExporterPushLogsData
--- FAIL: TestExporterPushLogsData (10.00s)
FAIL
FAIL    github.com/SigNoz/signoz-otel-collector/exporter/clickhouselogsexporter 10.376s
FAIL
```

tushar-signoz:
### Summary

`TestExporterPushLogsData` and `TestExporterPushTracesData` are both flaky, for the same reason.

Each expects the usage insert to run before the mock connection closes. But `stats.RecordWithTags` hands the measurement to the opencensus view worker **asynchronously**, and `worker.Read()` — what `usageCollector.Stop()` flushes through — reads the view directly without draining that queue. So the flush can find no usage rows, skip the insert, and leave the test failing on an expectation that was never met.

Both tests now wait for the view to report the data before calling `Shutdown`. `RetrieveData` round-trips through the worker channel, so it doubles as the barrier that guarantees the measurement has been applied.

### Related Issues
Closes #634

### Resource Impact

- [ ] Yes
- [x] No

Test-only.

### Testing

| | before | after |
|---|---|---|
| `clickhouselogsexporter` (full package) | 1 pass / 9 fail | 10 / 10 |
| `clickhousetracesexporter` (full package) | 4 pass / 1 fail | 10 / 10 |

Both pass when run alone, which is what made this confusing — and why the same tree could pass and fail CI on consecutive runs. The logs one reproduces reliably with the command from #634:

```bash
go test -timeout 30s -v -run 'TestExporterInit|TestExporterPushLogsData' ./exporter/clickhouselogsexporter -count=1
```

It also drops from 10.00s to 0.00s, since it no longer burns the full `eventually` timeout waiting for an insert that never comes. A full `go test ./...` sweep is clean.

### Note

`signozclickhousemetrics`'s `Test_shutdown` has the same setup — `ExpectExec` for the usage insert, then `Shutdown` — but never calls `ExpectationsWereMet()`, so the expectation can't fail and the test can't flake. Left alone here, but it isn't verifying the usage insert it appears to.

There's also a small production echo: usage recorded in the last moments before shutdown can be missed by the final flush, for the same reason. Not addressed.

### DIFF
--- a/exporter/clickhouselogsexporter/exporter_test.go
+++ b/exporter/clickhouselogsexporter/exporter_test.go
@@ -15,6 +15,7 @@ import (
 	cmock "github.com/srikanthccv/ClickHouse-go-mock"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.opencensus.io/stats/view"
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/otel/metric/noop"
@@ -157,6 +158,11 @@ func TestExporterPushLogsData(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to push logs data: %v", err)
 	}
+
+	eventually(t, func() bool {
+		rows, err := view.RetrieveData(SigNozLogsCount)
+		return err == nil && len(rows) > 0
+	})
 
 	err = exporter.Shutdown(context.Background())
 	assert.Nil(t, err)
--- a/exporter/clickhousetracesexporter/clickhouse_exporter_v3_test.go
+++ b/exporter/clickhousetracesexporter/clickhouse_exporter_v3_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/jellydator/ttlcache/v3"
 	cmock "github.com/srikanthccv/ClickHouse-go-mock"
 	"github.com/stretchr/testify/assert"
+	"go.opencensus.io/stats/view"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -768,6 +769,11 @@ func TestExporterPushTracesData(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to push traces data: %v", err)
 	}
+
+	eventually(t, func() bool {
+		rows, err := view.RetrieveData(SigNozSpansCount)
+		return err == nil && len(rows) > 0
+	})
 
 	err = exporter.Shutdown(context.Background())
 	assert.Nil(t, err)


.github/workflows/ci.yaml
```diff
@@ -1,6 +1,9 @@
 name: ci
 
 on:
+  push:
+    branches:
+      - main
```
krsoninikhil: We don't want to run the tests on feature branches?

tushar-signoz: Why to waste credits on feature branches

krsoninikhil: How would we know if a PR is breaking things, they will be detected only after merge.

tushar-signoz: CI will run on pushes on a pull requests. How it is already running.
This is just an additional merge time check

tushar-signoz: If I remove the main filter, CI would run for every commit pushed even without opening a PR, which I don't think we need right ?

krsoninikhil (dismissed):

.github/workflows/ci.yaml
krsoninikhil: Yeah, works

tushar-signoz:
🤣 Its still flaky

tushar-signoz:
Fix is correct, forgot adding it in traces

therealpandey (approved):

PR #891, opened 2026-08-24
fix(clickhouselogsexporter): faithful dual body ingestion via json_body_dual_ingestion

tushar-signoz:
### Summary

Dual ingestion writes each log body to both the legacy `body` column and the new `body_v2` JSON column, so reads can flip between the columns at any point of the migration with no data gap. The old dual-write flag couldn't deliver that: the `normalize` operator rewrites the body in place before the exporter sees it, so the "old" column actually received the normalized body — text lines stored as `{"message": ...}`, JSON strings re-serialized (key order, whitespace), `log`/`msg` keys renamed to `message`. Every `body` contains/LIKE query over dual-ingested data broke. Skipping normalize instead kept `body` faithful but degraded `body_v2` to `{"message": "<raw>"}` for everything.

This PR replaces `body_json_old_body_enabled` with a single `json_body_dual_ingestion` switch, present on both the normalize operator and the exporter and always set together (the SigNoz server will send both alongside its own flag of the same name — server-side change tracked separately):

**Normalize operator** (`json_body_dual_ingestion: true`, default false):
- Stashes the pre-normalization body on every record it processes, in the internal record attribute `__signoz_original_body__` — strings byte-exact, map bodies serialized before any mutation, scalars/bytes raw, nil as `""`.
- The attribute doubles as the per-record proof that normalize processed the log.

**ClickHouse logs exporter** (`json_body_dual_ingestion: true`, implies `body_json_enabled`):
- Restores the stash into the legacy `body` column, so it holds exactly what legacy ingestion would have stored.
- When dual ingestion alone enables body JSON, writes `body_v2` **only when the stash is present**: a window where the normalize operator isn't delivered leaves `body_v2` as `{}` instead of accumulating permanently wrapped `{"message": <raw>}` rows, while the body stays safe in the old column. Body *shape* is deliberately not used as the signal — user pipelines run after normalize and may legitimately stringify the body again; those records keep their stash and are wrapped into `body_v2`.
- With `body_json_enabled` set explicitly (reads are on `body_v2` in that regime), `body_v2` is always written, stash or not — an unwritten cell would be a hole in the queried column.
- Strips the stash attribute everywhere: attributes columns, tag/attribute-key tables, and usage size accounting. The metadata exporter skips it as well.

**Flag semantics** (`body_json_old_body_enabled` is removed; configs still carrying it fail validation loudly):

| `body_json_enabled` | `json_body_dual_ingestion` | `body` | `body_v2` |
| --- | --- | --- | --- |
| false | false | original | — |
| true | false | empty | always written (non-map bodies wrapped) |
| false | true | original (stash-restored) | written only for normalized records (stash present) |
| true | true | original (stash-restored) | always written |

Skew during a flip is safe in both directions: collector-first leaves `body_v2` temporarily `{}` (old `body` has everything); server-first still yields a faithful old `body` since the exporter restores the stash even with body JSON off. Ending the migration requires setting `body_json_enabled: true` in the same change that drops the dual flag, otherwise `body_v2` writes stop while queries still read it.

Design doc: [Dual JSON body ingestion](https://app.notion.com/p/signoz/Dual-JSON-body-ingestion-3c6fcc6bcd198049963bce6ce2648af8)

### Resource Impact
- [x] Yes
- [ ] No

If **Yes**, describe:
- **Resources impacted:** Memory (and network for gateway→worker topologies), only while `json_body_dual_ingestion` is on.
- **Expected change:** One extra record attribute carrying the original body through the pipeline: a string reference for string bodies (the dominant case, near-zero cost), one JSON serialization per structured map body. Dropped by the exporter before insert. Zero cost with the flag off — the default.


krsoninikhil:
> Every body contains/LIKE query over dual-ingested data broke.

Is this problem started after merging https://github.com/SigNoz/signoz-otel-collector/pull/887 or that is not related?

tushar-signoz:
> > Every body contains/LIKE query over dual-ingested data broke.
> 
> Is this problem started after merging #887 or that is not related?

Wording is not correct, It will break if we use the current dual ingestion, 887 is only a safety net and doesn't come into the picture unless user creates a bad pipeline, and is not related here

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -718,9 +720,12 @@ producerIteration:
 						e.logger.Warn("resourcemap exceeded the limit of 100 keys")
 					}
 					// record size calculation
-					attrBytes, _ := json.Marshal(record.Attributes().AsRaw())
+					attrsRaw := record.Attributes().AsRaw()
+					delete(attrsRaw, constants.OriginalBodyAttributeKey)
+					attrBytes, _ := json.Marshal(attrsRaw)
```
krsoninikhil: Not related to the change but since we are touching this line, lets add a warn log here. Marshaling can fail and looks like we won't even get to know currently.

exporter/clickhouselogsexporter/exporter_test.go
```diff
@@ -403,14 +419,14 @@ func TestProcessBody(t *testing.T) {
 				return v
 			},
 			expectedBody:     "42",
-			expectedBodyJSON: `{"message":"42"}`,
-			expectedPromoted: `{"message":"42"}`,
+			expectedBodyJSON: "{}",
+			expectedPromoted: "{}",
```
krsoninikhil: If both flags are enabled, expectedBodyJSON is empty json?

tushar-signoz: Fixed

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -836,27 +844,34 @@ producerIteration:
...
+			// promoted paths extraction using cached set
+			promotedSet := e.promotedPaths.Load().(map[string]struct{})
+			promoted = buildPromoted(bodyJSON, promotedSet)
+		}
 
-		if !e.bodyJSONOldBodyEnabled {
+		if !e.jsonBodyDualIngestion {
 			// set body to empty string
 			body = pcommon.NewValueEmpty()
+			hasOriginalBody = false
```
krsoninikhil: We are setting the bodyJSON if either of these values are true - JSONBodyEnabled, JSONBodyDualIngestion, hasOriginalBody

But body is set to empty if jsonBodyDualIngestion is false and JSONBodyEnabled is true?

3 knobs are controlling this behavior, let's see if we can get rid of dual ingestion flag considering we have the original body attribute also being used for writing to json. 

If original body attribute if set, then why can't we infer that dual ingestion is required?

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -836,27 +844,34 @@ producerIteration:
 	return nil
 }
 
-func (e *clickhouseLogsExporter) processBody(ctx context.Context, body pcommon.Value) (string, string, string) {
+func (e *clickhouseLogsExporter) processBody(ctx context.Context, body pcommon.Value, originalBody pcommon.Value, hasOriginalBody bool) (string, string, string) {
 	promoted := pcommon.NewValueMap()
 	bodyJSON := pcommon.NewValueMap()
 	if e.bodyJSONEnabled {
```
krsoninikhil, line 850: This will be true now even if just the dual ingestion is set to true. Is the expected?

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -328,17 +328,19 @@ func newExporter(_ exporter.Settings, cfg *Config, opts ...LogExporterOption) (*
 		maxAllowedDataAgeDays = *cfg.MaxAllowedDataAgeDays
 	}
 
+	bodyJSONEnabled := cfg.BodyJSONEnabled || cfg.JSONBodyDualIngestion
+
 	e := &clickhouseLogsExporter{
-		insertLogsSQLV2:           renderInsertLogsSQLV2(cfg.BodyJSONEnabled),
+		insertLogsSQLV2:           renderInsertLogsSQLV2(bodyJSONEnabled),
```
krsoninikhil, line 334: We have both flags in the exporter object, bodyJSONEnabled and jsonBodyDualIngestion. So overriding one with both values here seem confusing if just bodyJSONEnabled value is required later on.

krsoninikhil (changes requested):

exporter/clickhouselogsexporter/exporter.go
tushar-signoz: Didn't get you, bodyJSONEnabled is a variable

exporter/clickhouselogsexporter/exporter.go
tushar-signoz: yes

exporter/clickhouselogsexporter/exporter.go
tushar-signoz: > But body is set to empty if jsonBodyDualIngestion is false and JSONBodyEnabled is true?

Yes, this means that we have fully migrated a legacy tenant, or we are in the new ones. 

> If original body attribute if set, then why can't we infer that dual ingestion is required?

On the surface it sounds good, but the 3 knobs are not really 3 knobs, hasOriginalBody is a result of JSONBodyDualIngestion.
Now you may ask, what's wrong in only keeping hasOriginalBody and get rid of JSONBodyDualIngestion. JSONBodyDualIngestion is a policy and hasOriginalBody is a data fact. User can also forge this. Our situation allows to do this but its not the correct thing to do.

exporter/clickhouselogsexporter/exporter.go
```diff
@@ -836,27 +844,33 @@ producerIteration:
...
 	bodyJSON := pcommon.NewValueMap()
 	if e.bodyJSONEnabled {
-		if body.Type() == pcommon.ValueTypeMap {
-			// switch the reference to bodyJSON
-			bodyJSON = body
-		} else {
-			bodyJSON.Map().PutStr(bodyNonMapKey, getStringifiedBody(body))
-			e.nonMapBodyCounter.Add(ctx, 1)
-		}
+		writeBodyJSON := e.cfg.BodyJSONEnabled || hasOriginalBody
```
krsoninikhil, line 851: `e.bodyJSONEnabled` and `e.cfg.BodyJSONEnabled` can have different values currently. Let's not club both flags in `e.bodyJSONEnabled`, we can put a check for both on first if check here.

exporter/clickhouselogsexporter/exporter.go
krsoninikhil: It's also updating the flag value in the exporter https://github.com/SigNoz/signoz-otel-collector/pull/891/changes#diff-91bb2c3732348f626f9fbafd0b132217a5337e3e3b0efc0a9255f8abdc9517e5R337

exporter/clickhouselogsexporter/exporter.go
krsoninikhil: Fair

krsoninikhil (approved):
For sql template cleanup 🎉

PR #895, opened 2026-09-02
feat(clickhousetracesexporter): write calculated span fields as spanf…

srikanthccv:
…ield rows

https://github.com/SigNoz/signoz/pull/10110 was reverted because we have these calculated fields as quick filter fields, but they were not written to `tag_attributes_v2` so values was returning empty and reverted. This change populates so they can be used again.

- write `http_method`, `http_host`, `http_url`, `response_status_code`, `db_name`, `db_operation`, `external_http_method`, `external_http_url`, `is_remote` and `has_error` to `tag_attributes_v2` with `tag_type=spanfield`, next to the existing name/kind/status rows, so the fields API can suggest their values
- calculated rows are subject to the attribute value length limit and the high-cardinality skip list; intrinsic rows keep their always-write path
- dedupe span field rows per column and value within a batch instead of by value alone

Assisted-by: Claude Fable 5.1

### DIFF
--- a/exporter/clickhousetracesexporter/span_fields.go
+++ b/exporter/clickhousetracesexporter/span_fields.go
@@ -61,3 +61,33 @@ func appendSpanFieldString(rows []spanFieldRow, key, value string, calculated bo
 	}
 	return append(rows, spanFieldRow{key: key, dataType: utils.FieldDataTypeString, stringValue: value, calculated: calculated})
 }
+
+// calculatedSpanFieldColumns are the calculated columns and the data type they
+// are written with, which together form their entry in the skip list.
+var calculatedSpanFieldColumns = []struct {
+	key      string
+	dataType utils.FieldDataType
+}{
+	{key: "http_method", dataType: utils.FieldDataTypeString},
+	{key: "http_host", dataType: utils.FieldDataTypeString},
+	{key: "http_url", dataType: utils.FieldDataTypeString},
+	{key: "response_status_code", dataType: utils.FieldDataTypeString},
+	{key: "db_name", dataType: utils.FieldDataTypeString},
+	{key: "db_operation", dataType: utils.FieldDataTypeString},
+	{key: "external_http_method", dataType: utils.FieldDataTypeString},
+	{key: "external_http_url", dataType: utils.FieldDataTypeString},
+	{key: "is_remote", dataType: utils.FieldDataTypeString},
+	{key: "has_error", dataType: utils.FieldDataTypeBool},
+}
+
+// skippedSpanFieldColumns resolves the skip list for the calculated columns
+// once per batch, so the per-row check is a lookup by column name.
+func skippedSpanFieldColumns(shouldSkipKeys map[string]shouldSkipKey) map[string]struct{} {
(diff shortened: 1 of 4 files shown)


exporter/clickhousetracesexporter/writer.go
```diff
@@ -356,7 +356,9 @@ func (w *SpanWriter) writeTagBatchV3(ctx context.Context, batchSpans []*SpanV3)
 	// create map of span attributes of key, tagType, dataType, isColumn and value to avoid duplicates in batch
 	mapOfSpanAttributeValues := make(map[string]struct{})
 
-	mapOfSpanFields := make(map[string]struct{})
+	mapOfSpanFields := make(map[spanFieldDedupeKey]struct{})
+	spanFields := make([]spanFieldRow, 0, 16)
```
krsoninikhil, line 360: Size of map above will also have a same max value?

exporter/clickhousetracesexporter/span_fields.go
```diff
@@ -0,0 +1,93 @@
...
+	{key: "http_host", dataType: utils.FieldDataTypeString},
+	{key: "http_url", dataType: utils.FieldDataTypeString},
+	{key: "response_status_code", dataType: utils.FieldDataTypeString},
+	{key: "db_name", dataType: utils.FieldDataTypeString},
+	{key: "db_operation", dataType: utils.FieldDataTypeString},
+	{key: "external_http_method", dataType: utils.FieldDataTypeString},
+	{key: "external_http_url", dataType: utils.FieldDataTypeString},
+	{key: "is_remote", dataType: utils.FieldDataTypeString},
+	{key: "has_error", dataType: utils.FieldDataTypeBool},
+}
```
krsoninikhil, line 81: If this can be a map of calculated fields, then all the rows in `spanFieldRows` can also use it as a source to define calculated fields instead of having that knowledge baked in. 

Would these many map lookup on every span be costly enough to repeat this calculated field information?

krsoninikhil (approved):

exporter/clickhousetracesexporter/span_fields.go
```diff
@@ -0,0 +1,93 @@
...
+
+// spanFieldDedupeKey identifies a row within a write batch. It is a struct so
+// that map lookups do not allocate.
+type spanFieldDedupeKey struct {
+	key         string
+	dataType    utils.FieldDataType
+	stringValue string
+	numberValue float64
+}
```
krsoninikhil, line 29: If `calculatedSpanFieldColumns` becomes a map, then these two struct can be merged to one as well since any check for `calculated` field can be a map lookup.

exporter/clickhousetracesexporter/writer.go
srikanthccv: no, the map holds an uniqu dedup for entire batch, but the spanFields is a reuable slice of 16 top level or intrinsic fields rows for each span

exporter/clickhousetracesexporter/span_fields.go
srikanthccv: yes

PR #906, opened 2026-09-14
chore(pkg/tagdedup): shared per-batch dedup for clickhouse tag table writes

therealpandey:
## Summary

Extracts the map-based dedup logic introduced for spans in #177 into a shared package (`pkg/tagdedup`) so it can be used by both the traces and logs exporters without duplication.

This PR **only adds the package** — no exporter changes. Adoption will follow in two separate PRs:
1. logs exporter: add per-batch dedup of tag table writes (implements #177 for logs)
2. traces exporter: replace the inline maps in `writeTagBatchV3` with this package

## What's in the package

- `KeyID(key, tagType, dataType, isColumn)` — dedup ID for attribute key rows
- `ValueID(key, tagType, dataType, stringValue, numberValue)` — dedup ID for attribute value rows
- `Deduper` — per-batch tracker with `SeenKeyID`/`SeenValueID` (first call records and returns false, subsequent calls return true)

IDs are built with a unit-separator delimiter to avoid the concatenation collisions possible in the original inline implementation.

## Testing

- Unit tests cover ID determinism, distinctness across every component, collision resistance, float formatting, and key/value space independence.
- `go test ./pkg/tagdedup/...` and `go vet ./pkg/tagdedup/...` pass.


tushar-signoz (approved):
LGTM

PR #907, opened 2026-09-14
chore(clickhouselogsexporter): use pkg/tagdedup to reduce tag table writes

therealpandey:
Stacked on #906. Implements https://github.com/SigNoz/signoz-otel-collector/pull/177 for logs (ref https://github.com/SigNoz/engineering-pod/issues/1028).

## Summary

Uses the shared `pkg/tagdedup` Deduper to skip duplicate rows written to the logs tag tables within a batch:

- one `Deduper` per `pushToClickhouse` batch, used only from the single consumer goroutine (no concurrency concerns)
- `addAttrsToAttributeKeysStatement` returns early when `SeenKeyID` reports the (key, tagType, dataType) row already written this batch — layered on top of the existing cross-batch TTL `keysCache`
- `addAttrsToTagStatement` skips the `tag_attributes_v2` append when `SeenValueID` reports the (key, tagType, dataType, value) row already written this batch, matching the traces behavior in `writeTagBatchV3`

This also dedupes `logfield` rows (severity_text/severity_number), which repeat on every record.

## Testing

- `go build`, `go vet`, `gofmt` clean
- `go test ./exporter/clickhouselogsexporter/...` passes


tushar-signoz (dismissed):
LGTM

krsoninikhil (approved):

PR #913, opened 2026-09-17
chore: don't bill signoz gen ai attributes

nityanandagohain:
### Summary
* Don't bill the attributes generated for gen_ai cost calculation
* Drop `signoz.gen_ai` attributes if they are sent by the user
* Standardize the attribute names from https://github.com/SigNoz/signoz-semantic-conventions/pull/7/changes

### Related Issues

Closes https://github.com/SigNoz/engineering-pod/issues/4774

### Resource Impact
- [ ] Yes
- [x] No


exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -284,6 +284,28 @@ func (attrMap *attributesData) add(key string, value pcommon.Value) {
 	attrMap.SpanAttributes = append(attrMap.SpanAttributes, spanAttribute)
 }
 
+// billableNumberAttributes returns attrs without LLM pricing costs. Most spans
+// carry none, so the input map is returned as is and only copied when a cost key is present.
+func billableNumberAttributes(attrs map[string]float64) map[string]float64 {
+	hasExcluded := false
+	for k := range attrs {
+		if metering.ExcludeSigNozLLMPricingSpanAttrs.MatchString(k) {
```
nityanandagohain, line 292: it's expensive but will have to do it since we have not shifted entirely to meter for billing

therealpandey (approved):

exporter/clickhousetracesexporter/clickhouse_exporter_v3.go
```diff
@@ -284,6 +284,28 @@ func (attrMap *attributesData) add(key string, value pcommon.Value) {
 	attrMap.SpanAttributes = append(attrMap.SpanAttributes, spanAttribute)
 }
 
+// billableNumberAttributes returns attrs without LLM pricing costs. Most spans
+// carry none, so the input map is returned as is and only copied when a cost key is present.
+func billableNumberAttributes(attrs map[string]float64) map[string]float64 {
+	hasExcluded := false
+	for k := range attrs {
```
vikrantgupta25, line 291: you don't need two loops for this O(2*attr) can be brought down to O(attr)

vikrantgupta25: since you are looping all over the attrs then just build the billableattrs here itself than break

nityanandagohain: but then I will have to allocate memory all the times, so the chance that hasExcluded is true is <10% and only for them it will be O(2*attr)

vikrantgupta25: you are right, makes sense

vikrantgupta25 (approved):

